PR #30, opened 2016-05-05
chore(style): Added editorconfig, Added tslint as part of npm script

divyakumarjain:
chore(style): Added editorconfig, Added tslint as part of npm script
Breaking changes: npm run flow.compile will fail if linting of typescript file fails


mohammedzamakhan:
@divyakumarjain Can you remove the conflicts and commit please.

Thanks

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
```
divyakumarjain: Noticed that not all the files are consistently using indent_size 2 some are using 4. Should it be 2 or 4.

mohammedzamakhan: It should be 2

PR #36, opened 2016-05-10
feat(core): Implemented support for hideExpression

divyakumarjain:
Implemented support for hideExpression as String expression, boolean and function

#14


divyakumarjain:
Working sample for hideExpression. Implementation can be done in different way where hideExpression can bind to [hidden]. Tried but it did not work for some reason.

mohammedzamakhan:
Cool, Can you resolve the conflict and push again?

Thanks

src/services/formly.field.visibility.ts
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+    hasHideExpression() :boolean {
+        return this.formlyField.field.hideExpression !== undefined && this.formlyField.field.hideExpression
+    }
+
+    checkVisibilityChange() {
+        var hideExpressionResult: boolean = this.eval(this.formlyField.field.hideExpression);
+        if (this.formlyField.field.hideExpression !== undefined && hideExpressionResult !== this.formlyField.isHidden())  {
+
```
mohammedzamakhan: Remove line

divyakumarjain: I have removed redundant check for hideExpression. But check for change in hideExpressionResult with respect to formlyfield.hide ensures that setHidden is only called when hideExpressionResult has actually changed.

Issue #12, opened 2016-04-13
'focus' can only be true for one field


mohammedzamakhan:

divyakumarjain:
Updated the pull request with latest from master.

PR #38, opened 2016-05-15
Remove Control Service

divyakumarjain:
Apart from removing control service. This pull request will also help #13 and #15.


divyakumarjain:
Please also look at comments for individual commits.

src/services/formly.processor.ts
```diff
@@ -0,0 +1,23 @@
...
+      this.visitors.forEach(visitor => {
+        visitor.visit(field);
+      });
+    });
+  }
+}
+
+class FormlyConfigValidator implements  FormlyConfigVisitor {
+  visit (field: FormlyFieldConfig) {}
+}
```
divyakumarjain: Wanted to use Visitor pattern to create control group and also address #12 by having visitor also validate the config. But In final solution this was not used as found that individual template directive should hand control group creation. Left this implementation for future use of while addressing #12.

Issue #24, opened 2016-05-04
Not working expressionProperties 


Mig1st4ck:
In the current demo the expressionProperties don't work.

Also expressionProperties should work with both expressions (strings) and (functions) same as the original formly.

PR #40, opened 2016-05-17
feat(expressionProperties): implemented expressionProperties

divyakumarjain:
implemented expressionProperties expression evaluation and setting the expression value back to formly field

This will address #24


demo/index.html
```diff
@@ -84,8 +84,9 @@
 </body>
 
 <!-- ES6-related imports -->
-<script src="/node_modules/angular2/bundles/angular2-polyfills.js"></script>
 <script src="/node_modules/es6-shim/es6-shim.min.js"></script>
+<script src="/node_modules/zone.js/dist/zone.js"></script>
+<script src="/node_modules/reflect-metadata/Reflect.js"></script>
 <script src="/node_modules/systemjs/dist/system.js"></script>
```
divyakumarjain: After moving to Angular RC and cleaning up node_module was getting issue in browser. Adding zone.js and Reflect.js resolves the problem.

demo/index.html
```diff
@@ -128,11 +129,10 @@
   }});
 </script>
 <script src="/node_modules/rxjs/bundles/Rx.js"></script>
-<script src="/node_modules/angular2/bundles/angular2.js"></script>
 <script src="../bundles/ng2-formly.js"></script>
```
divyakumarjain: As mention earlier demo was pointing to angular2 instead of @angular. The clean up of node_module and npm install again will help resolve any problem.

PR #50, opened 2016-05-29
fix(formBinding): Fix for two way binding issue

divyakumarjain:
This Commits provides a temporary fix for two binding issue (see#45 for more details). This workaround for observable change propagation needs to be removed once issue with angular two way binding for dynamic  component is fixed. See angular/angular#8903 for more.

This Pull request also has refactoring changes.

### DIFF
--- a/demo/template.html
+++ b/demo/template.html
@@ -11,6 +11,7 @@
     <br><strong>FORM DATA:</strong><br><pre>{{user | json}}</pre><br>
     <hr>
     <button (click)="showEmail()" class="btn btn-success">Change Email</button>
+    <button (click)="resetForm()" class="btn btn-danger">Reset</button>
     <button (click)="hide()" class="btn btn-info">Hide Expression Toggle</button>
 
 
--- a/demo/template.html
+++ b/demo/template.html
@@ -2,7 +2,7 @@
     <h1>NG2-formly example: Introduction</h1>
     <p>This is a small subset of the things that formly can do :-) See the other examples and the documentation for more.</p>
     <hr>
-    <formly-form [viewModel]="user" [fields]="userFields" (formSubmit)="console($event)" [form]="form">
+    <formly-form [model]="user" [fields]="userFields" (formSubmit)="console($event)" [form]="form">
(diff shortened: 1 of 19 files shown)


demo/template.html
```diff
@@ -2,7 +2,7 @@
     <h1>NG2-formly example: Introduction</h1>
     <p>This is a small subset of the things that formly can do :-) See the other examples and the documentation for more.</p>
     <hr>
-    <formly-form [model]="user" [fields]="userFields" (formSubmit)="console($event)" [form]="form">
+    <formly-form [viewModel]="user" [fields]="userFields" (formSubmit)="console($event)" [form]="form">
```
mohammedzamakhan: Why do you think this has to be changed?

divyakumarjain: This was changed because viewModel is now defined as part of FormlyCommon. If we don't want to change public API we have two options.

1. Update FormlyCommon to use model instead of viewModel. But I feel view Model is correct convention for FormlyField and FormlyGroups as it track state of view. Also I agree model would be right convention for FormlyForm.

2. Use Model for public API but internal track state within viewModel.

I will spend some time on this and will update PR. But do let me know what everybody thinks. I am open for suggestions.

divyakumarjain:
Please Do look at following items too.

1. FormlyGroup component: I am thinking to use that to group fields on UI and also it can be used in scenarios where FieldGroup has a key. For E.g. all the address related field like Street, City, Pin  needs to group on UI and also at Data Model level.

2. FormlyFieldBuilder: This is used to move out logic of creating child template directive dynamically. This also Simplifies  FormlyField. Still few event listeners are part of FormlyField which needs to be moved to FromlyFieldBuilder. But will take that in future PR.

Once this stabilizes will raise smaller PRs. At present I understand this PR has bug fix and refactoring both.

divyakumarjain:
@mohammedzamakhan i have updated the PR and has replaced viewModel with model. At Present two way binding for Radio button is not working. I guess its a bug in angular where Radio button and ngModel they do not go well. I am still investigating on that and will raise new PR for fixing 2-way binding for Radio button.

PR #60, opened 2016-06-18
fix(radiobutton): Fixes #54 issue with two way binding of radio button

### LINKED ISSUE #54: Two Way binding for Radio button does not work
Two binding issue discussed in https://github.com/formly-js/ng2-formly/issues/45 was fix for most of the component except for radio button.

When [(ngModel)] is added to the Radio button then the RadioButtonValueAccessor throws an error that _state object is undefined when radio button is switched. This seem like a bug in Angular. We will create a small plunk sample and share it on Angular github and stackoverflow.

divyakumarjain:
Fix for two way binding of radio button mentioned in #54

### DIFF
--- a/src/components/formly.common.component.ts
+++ b/src/components/formly.common.component.ts
@@ -1,4 +1,4 @@
-import {Input, Output, EventEmitter, ElementRef, OnChanges} from "@angular/core";
+import {Input, Output, EventEmitter, ElementRef, Renderer} from "@angular/core";
 import {FormlyFieldExpressionDelegate, FormlyFieldVisibilityDelegate} from "../services/formly.field.delegates";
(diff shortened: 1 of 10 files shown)


src/components/formly.common.component.ts
```diff
@@ -36,19 +37,20 @@ export class FormlyCommon {
...
+  @Input()
+  public get hide() {
+    return this._hide;
   }
-  setHidden(cond: boolean) {
-    this.hide = cond;
+  public set hide(value: boolean) {
+    this._hide = value;
 
-    this.elem.nativeElement.style.display = cond ? "none" : "";
```
divyakumarjain: Please review if this is best way to use renderer so that it can run on server side.

PR #85, opened 2016-08-20
fix(style): fix #82 by adding form-control & form-group css classes where missing

### LINKED ISSUE #82: css class form-control missing in most templates
input and textearea templates both have the css class form-control included. All others don't. This causes styling to be non-uniform between different fields.

divyakumarjain:
fix #82 by adding form-control in checkbox, select, radio and multicheckbox templates also added form-group in select template.

### DIFF
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -1,0 +1,13 @@
+# Unreleased
+
+### Bug Fixes
+* Add missing form-control css class to select bootstrap template ([#82](https://github.com/formly-js/ng2-formly/issues/82))([pull request 85](https://github.com/formly-js/ng2-formly/pull/85))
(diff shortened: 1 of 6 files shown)


mohammedzamakhan:
The templates were based off bootstrap 4 for the checkbox and radio box (http://v4-alpha.getbootstrap.com/components/forms/#custom-forms)

Even bootstrap 4 has changed the classes for custom inputs, Can you check whether this would look fine with bootstrap 3 and bootstrap 4 both css files?

divyakumarjain:
the PR only adds wrapper empty css class on input elements and thus should not impact custom inputs unless there are global css styles using these wrapper classes and messing up the custom inputs, I dont see any issue. These empty css classes can be used for application wide styling of input controls within forms.

But still I am personally interested in testing it ng2-formly against bootstrap 4  to understand how they go together.

Also going forward once we have separate modules for each template implementation. it would be concern of individual template implementation to worry about such classes.

divyakumarjain:
Have tested against bootstrap 3 & 4 and have updated classes for bootstrap 4.

Also seems that form-control class on radio and checkbox does not go well with bootstrap 3. Thus reverted those changes.

PR #87, opened 2016-08-25
separate bootstrap from the core package.

aitboudad:
selection_025


divyakumarjain:
@mohammedzamakhan Do you want to review this pull request. I think there was a plan to have separate git repo for templates.  

Also we can think of monorepo the way angular, material2 are doing it. May be this PR is step in that direction.

divyakumarjain:
@aitboudad can you please consider using git-cz to commit files it generates better commit messages.

mohammedzamakhan:
@divyakumarjain Yes I think this is the step towards our goal of monorepo.

Eventually we can have all bootstrap templates in this repo

https://github.com/formly-js/ng2-formly-templates-bootstrap

aitboudad:
Hi 
I would prefer to keep it in the same repo (like angular does), at least right now we can keep it as it is until we achieve `v1.0` WDYT ?

PR #93, opened 2016-09-18
feat(formly-form): move `form` out of `formly-form` component.

aitboudad:
Hi,
I would like to use `ngSubmit` and I figured out that is not possible with the current implementation, so I propose to not include `form` in the `formly-form` to allow more flexibility.

thanks in advance :smile:


mohammedzamakhan:
@aitboudad update the readme file. Thanks

aitboudad:
done #98

PR #99, opened 2016-09-23
fix(validation): validation occurs before the user interacts.

aitboudad:
Note: I also updated angular to v2.0 except `@angular/forms`


mohammedzamakhan:
Looks good to me except the failure 😄

mohammedzamakhan:
Note is not valid anymore 🌵

PR #110, opened 2016-09-29
feat(validation) Support for Validation Array fixes #41

### LINKED ISSUE #41: Support for Validation to be An Array of strings

mohammedzamakhan:
This will make sure that Validations can be in an array, and thus JSON can be used to populate the form with validations.

### DIFF
--- a/src/components/formly.field.ts
+++ b/src/components/formly.field.ts
@@ -2,6 +2,7 @@ import {
   Component, OnInit, EventEmitter, ElementRef, Output,
   ViewContainerRef, ViewChild, ComponentRef, Renderer
 } from "@angular/core";
+import {Validators} from "@angular/forms";
 import {FormlyCommon} from "./formly.common.component";
 import {FormlyPubSub, FormlyEventEmitter, FormlyValueChangeEvent} from "../services/formly.event.emitter";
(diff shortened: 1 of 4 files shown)


src/templates/field.ts
```diff
@@ -24,7 +27,14 @@ export abstract class Field implements OnInit {
 
   get formControl(): AbstractControl {
     if (!this._control) {
-     this.createControl();
+      if (typeof this.field.validation === "object") {
```
aitboudad: I would prefer to move this into `FormlyField` component

mohammedzamakhan: Makes sense. Thanks 😄

mohammedzamakhan: Done 💯

src/components/formly.field.ts
```diff
@@ -55,6 +56,13 @@ export class FormlyField extends FormlyCommon implements OnInit {
 
   createChildFields() {
     if (this.field && !this.field.template && !this.field.fieldGroup) {
+      if (typeof this.field.validation === "object") {
```
aitboudad: can we use `Array.isArray(this.field.validation)` instead ?

mohammedzamakhan: Done 💯

PR #112, opened 2016-10-01
fix(changedetection): remove the temporary fix `setTimeout`.

aitboudad:


aitboudad:
@mohammedzamakhan can I merge :) ?

src/templates/formlyfield.multicheckbox.ts
```diff
@@ -23,15 +24,12 @@ import {Field} from "./field";
...
-  constructor(private formBuilder: FormBuilder) {
-    super();
-  }
-
-  createControl(): AbstractControl {
-    let controlGroupConfig = this.templateOptions.options.reduce((previous, option) => {
-      previous[option.key] = [this.model ? this.model[option.key] : undefined];
+  static createControl(model: any, field: FormlyFieldConfig): AbstractControl {
+    let controlGroupConfig = field.templateOptions.options.reduce((previous, option) => {
+      previous[option.key] = new FormControl(model ? model[option.key] : undefined);
```
divyakumarjain: Should we document this contract that if any template needs to  create custom controls they can provide this static method.

At present I am not sure where this can be documented :) thus this comment should not be showstopper for this PR.

src/services/formly.providers.ts
```diff
@@ -1,7 +1,9 @@
 import {FormlyPubSub} from "./formly.event.emitter";
 import {FormlyMessages} from "./formly.messages";
+import {SingleFocusDispatcher} from "./formly.single.focus.dispatcher";
 
 export const FormlyProviders = [
   FormlyPubSub,
-  FormlyMessages
+  FormlyMessages,
+  SingleFocusDispatcher
```
divyakumarjain: I kind of agree that SingleFocusDispatcher should be a global level not at individual formly level.

But will global SingleFocusDispatcher cause any problem. if not this looks good.

src/components/formly.form.ts
```diff
@@ -42,9 +41,36 @@ export class FormlyForm extends FormlyCommon implements OnInit  {
...
+        } else {
+          this.form.addControl(field.key, new FormControl({ value: this.model[field.key] || "", disabled: field.templateOptions.disabled }, field.validation));
+        }
+      }
+
+      if (field.fieldGroup) {
+        this.registerFormControls(field.fieldGroup);
+      }
+    });
+  }
```
divyakumarjain: Will there be a problem when new field configuration is dynamically added later (i.e. after ngInit) ? Not sure if original formy supports this.

src/services/formly.providers.ts
aitboudad: keep it private is better!

src/components/formly.form.ts
aitboudad: yes and that's the reason of moving it to the `FormlyForm`

src/templates/formlyfield.multicheckbox.ts
aitboudad: just opened an issue #115

PR #129, opened 2016-10-05
feat(formly-field) support extending types (with docs)

mohammedzamakhan:


src/components/formly.field.ts
```diff
@@ -115,7 +115,9 @@ export class FormlyField implements DoCheck, OnInit {
         fieldComponent = wrapperRef.instance.fieldComponent;
       });
     }
-
+    if (!type.component) {
```
aitboudad: instead it would better to handle this case inside `getType` method!

PR #141, opened 2016-10-18
fix(#140): assign default templateOptions.

### LINKED ISSUE #140: Placeholders are not optional (display 'undefined')
Using Formly 2.0.0-beta.7.

Placeholders are optional in templateOptions (and that makes sense, you potentially don't need them), but not setting a placeholder results in input elements using the string 'undefined' as a placeholder.
```
{
    "key" : "inhalt",
    "type" : "input",
    "templateOptions" : {
        "label" : "Inhalt"
        // no placeholder
    },
    "className" : "col-xs-3"
}
```
formly_undefined

aitboudad:
fix #140

### DIFF
--- a/src/templates/wrappers/description.ts
+++ b/src/templates/wrappers/description.ts
@@ -11,5 +11,5 @@ import {FieldWrapper} from '../field.wrapper';
   `,
 })
 export class FormlyWrapperDescription extends FieldWrapper {
-  @ViewChild("fieldComponent", {read: ViewContainerRef}) fieldComponent: ViewContainerRef;
+  @ViewChild('fieldComponent', {read: ViewContainerRef}) fieldComponent: ViewContainerRef;
 }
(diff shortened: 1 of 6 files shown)


src/templates/wrappers/description.ts
```diff
@@ -11,5 +11,5 @@ import {FieldWrapper} from '../field.wrapper';
   `,
 })
 export class FormlyWrapperDescription extends FieldWrapper {
-  @ViewChild("fieldComponent", {read: ViewContainerRef}) fieldComponent: ViewContainerRef;
+  @ViewChild('fieldComponent', {read: ViewContainerRef}) fieldComponent: ViewContainerRef;
 }
```
divyakumarjain: @mohammedzamakhan @aitboudad do we plan to use single quote or double quote? The rules in codacy and tslint.json may not be consistent.

aitboudad: `single quote` to be consistent with angular

divyakumarjain: @mohammedzamakhan should we update codacy if its not consistent. Also codacy support organization so that you can add aitboudad and me.

Issue #121, opened 2016-10-03
Nested Property Keys


mohammedzamakhan:
INFO: http://angular-formly.com/#/example/advanced/nested-property-keys

mohammedzamakhan:
@divyakumarjain @aitboudad Should we support this? since Angular 2 itself is not supporting it.

divyakumarjain:
For scenarios where form layout is not nested but model does have nested properties it will be tricky to place nested fields across the form layout. But keeping these fields grouped at single location in layout is possible using formly groups.

i think is good to have but may not be must have.

aitboudad:
I did some initial work and seems to be working fine, I'll provide a PR soon!

mohammedzamakhan:
Thanks @Pouja 🎆 🌟 🎉

FloNeu:
Ok. Hi, again guys - so this actually should be supported... But i am running into problems with this.
I got a nested model and want to refer to one of the keys from my FormlyfieldConfig, like: `key: "district.id"` with a model like`{ district: { id: 'UniqueId" }}` but when i set my model on the form the nested field value just isn't reflected (/gets deleted) on the model anymore.

PR #142, opened 2016-10-18
fix(#121): support nested keys

Pouja:
Nested keys were supported to some degree as in the model would update
but it would not properly set the form control.

This was due the fact that the formgroup would not nest when a
fieldgroup was specified.

This was fixed by making registerFormControls context sensitive in the
sense that the nested model and nested formgroup will be passed if in
the formfieldconfig the key was specified.

Updated the demo to showcase that indeed the value will be set at the
start.

### DIFF
--- a/src/components/formly.form.ts
+++ b/src/components/formly.form.ts
@@ -38,14 +38,14 @@ export class FormlyForm implements OnInit  {
       this.form = this.formBuilder.group({});
     }
 
-    this.registerFormControls(this.fields, this.form);
+    this.registerFormControls(this.fields, this.form, this.model);
   }
 
   changeModel(event: FormlyValueChangeEvent) {
     this.model[event.field.key] = event.value;
   }
 
-  private registerFormControls(fields, form) {
+  private registerFormControls(fields, form, model) {
     fields.map(field => {
       if (field.key && field.type) {
         field.templateOptions = Object.assign({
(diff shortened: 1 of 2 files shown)


src/components/formly.form.ts
```diff
@@ -63,18 +63,21 @@ export class FormlyForm implements OnInit  {
...
         } else {
-          form.addControl(field.key, new FormControl({ value: this.model[field.key] || '', disabled: field.templateOptions.disabled }, field.validation));
+          form.addControl(field.key, new FormControl({ value: model[field.key] || '', disabled: field.templateOptions.disabled }, field.validation));
         }
       }
 
       if (field.fieldGroup) {
         if (field.key) {
-          form.addControl(field.key, new FormGroup({}, field.validation));
+          const nestedForm = new FormGroup({}, field.validation);
```
aitboudad: you shoud pass nestedModel in `FormGroup`
```ts
const nestedModel = model[field.key] || {};
const nestedForm = new FormGroup(nestedModel, field.validation);
```

Pouja: May I ask why? Because `nestedModel` is just a arbitrary object, whereas the first argument of `FormGroup` accepts only `FormControls` (`AbstractControl` to be more precise).

aitboudad: sorry I was confused, it should be `nestedForm.setValue(nestedModel);`

aitboudad: or leave it as it is, I'll provide another PR to simplify passing model values!

aitboudad:
Thank you @Pouja

Pouja:
@aitboudad No problem!

PR #151, opened 2016-10-21
feat(Lifecycle Hooks) Integrate the fields with lifecycle hooks

### LINKED ISSUE #150: Lifecycle Hooks in Formly
We should have control over lifecycle for each fields. My idea is to implement something of this nature

```ts
      {
          key: 'email',
          type: 'input',
          templateOptions: {
            type: 'email',
            label: 'Email address'
          },
          lifecycle: {
            onInit: function() {
              console.log('on Init function', this);
              this.templateOptions.label = 'Email Address Updated';
            },
            onChanges: function() {
              console.log('on changes function', this);
            },
            doCheck: function() {
              console.log('on docheck function', this);
            },
            afterContentInit: function() {
              console.log('on afterContentInit function', this);
            }
          }
        }
```

mohammedzamakhan:
fixes #150

### DIFF
--- a/src/templates/field.type.ts
+++ b/src/templates/field.type.ts
@@ -1,0 +1,46 @@
+import {Field} from './field';
+import { OnInit, OnChanges, DoCheck, AfterContentInit, AfterContentChecked, AfterViewInit, OnDestroy, AfterViewChecked } from '@angular/core';
+
+export abstract class FieldType extends Field implements OnInit, OnChanges, DoCheck, AfterContentInit, AfterContentChecked, AfterViewInit, AfterViewChecked, OnDestroy {
+
+  private get lifecycle() {
+    return this.field.lifecycle;
+  }
+  ngOnInit() {
+    this.lifeCycleHooks('onInit');
+  }
+
+  ngOnChanges(changes) {
+    this.lifeCycleHooks('onChanges');
+  }
+
+  ngDoCheck() {
+    this.lifeCycleHooks('doCheck');
+  }
+
+  ngAfterContentInit() {
+    this.lifeCycleHooks('afterContentInit');
+  }
+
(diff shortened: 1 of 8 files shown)


src/components/formly.field.ts
```diff
@@ -130,6 +130,7 @@ export class FormlyField implements DoCheck, OnInit {
         form: this.form,
         field: this.field,
         formModel: this.formModel,
+        lifecycle: this.field.lifecycle,
```
aitboudad: we can avoid this by adding `lifecycle` getter in the `FieldType` class
```ts
get lifecycle() { return this.field.lifecycle; }
```

src/templates/field.type.ts
```diff
@@ -0,0 +1,53 @@
+import {Field} from './field';
+import { Input, OnInit, OnChanges, DoCheck, AfterContentInit, AfterContentChecked, AfterViewInit, OnDestroy, AfterViewChecked } from '@angular/core';
+
+export abstract class FieldType extends Field implements OnInit, OnChanges, DoCheck, AfterContentInit, AfterContentChecked, AfterViewInit, AfterViewChecked, OnDestroy {
+  @Input() lifecycle: any;
+  ngOnInit() {
+    if (this.lifecycle && this.lifecycle.onInit) {
```
aitboudad: I would prefer to move this into a private method something like:
```ts
  ngOnInit() {
    this.lifecycleHook('onInit');
  }
  private lifecycleHook(type, params = []) {
    if (this.lifecycle && this.lifecycle[type]) {
      this.lifecycle[type].bind(this)(...params);
    }
  }
```

FloNeu:
Can somebody point to an example of this being used? I couldn't find anything in the documentation... Given that i want to wait until formly is done setting up the form...

PR #156, opened 2016-10-24
feat(Default Value) Add default Value functionility

### LINKED ISSUE #154: defaultValue property for a field
We should support defaultValue attribute similar to angular formly 1

http://docs.angular-formly.com/v6.3.2/docs/field-configuration-object#defaultvalue-any

mohammedzamakhan:
fixes #154

### DIFF
--- a/src/components/formly.field.ts
+++ b/src/components/formly.field.ts
@@ -56,7 +56,7 @@ export class FormlyField implements DoCheck, OnInit {
     private formlyPubSub: FormlyPubSub,
     private renderer: Renderer,
     private formlyConfig: FormlyConfig,
-    private componentFactoryResolver: ComponentFactoryResolver
+    private componentFactoryResolver: ComponentFactoryResolver,
   ) {}
 
   ngDoCheck() {
--- a/src/components/formly.form.ts
+++ b/src/components/formly.form.ts
@@ -21,6 +21,7 @@ export class FormlyForm implements OnInit, OnChanges {
   @Input() model: any = {};
   @Input() form: FormGroup = new FormGroup({});
   @Input() fields: FormlyFieldConfig[] = [];
+  private defaultPath;
 
   constructor(private formlyConfig: FormlyConfig) {}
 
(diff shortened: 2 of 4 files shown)


src/components/formly.field.ts
```diff
@@ -56,7 +56,7 @@ export class FormlyField implements DoCheck, OnInit {
     private formlyPubSub: FormlyPubSub,
     private renderer: Renderer,
     private formlyConfig: FormlyConfig,
-    private componentFactoryResolver: ComponentFactoryResolver
+    private componentFactoryResolver: ComponentFactoryResolver,
```
aitboudad: extra comma

src/components/formly.field.ts
```diff
@@ -66,6 +66,7 @@ export class FormlyField implements DoCheck, OnInit {
 
   ngOnInit() {
     this.createChildFields();
+    this.checkDefaultValue();
```
aitboudad: why we need to check DefaultValue here, checking in [FormlyForm::addFormControl](https://github.com/formly-js/ng2-formly/blob/master/src/components/formly.form.ts#L110) is enough I think!

src/components/formly.form.ts
```diff
@@ -24,7 +24,7 @@ export class FormlyForm implements OnInit  {
 
   constructor(
     private formlyConfig: FormlyConfig,
-    private formBuilder: FormBuilder
+    private formBuilder: FormBuilder,
```
aitboudad: extra comma

mohammedzamakhan: Trailing comma's have become a practice in the industry, so that you do not have two line changes.

I would keep this change, in fact we can make this mandatory in tsconfig as well

aitboudad: you means `tslint` right!

src/components/formly.field.ts
mohammedzamakhan: I wanted field.key of nested.arrays.0, but in FormlyForm, I was only getting key as 0, so I decided to have it here in FormlyField

PR #164, opened 2016-10-25
feat(Validation) add Validators API

### LINKED ISSUE #163: Simpler Validation
We should have the angular-formly validators API in ng2-formly as well.

ref: http://angular-formly.com/#/example/advanced/validators

mohammedzamakhan:
fixes #163

### DIFF
--- a/demo/hello.ts
+++ b/demo/hello.ts
@@ -4,7 +4,7 @@ import 'core-js/es7/reflect';
 import 'zone.js/dist/zone';
 
 import {NgModule, Component, ViewChild, ViewContainerRef} from '@angular/core';
-import {FormsModule, ReactiveFormsModule, Validators, FormBuilder, FormGroup} from '@angular/forms';
+import {FormsModule, ReactiveFormsModule, Validators, FormBuilder, FormGroup, FormControl} from '@angular/forms';
 import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
 import {BrowserModule} from '@angular/platform-browser';
 import {FormlyModule, FormlyFieldConfig, FormlyBootstrapModule, Field, FieldWrapper} from './../src/index';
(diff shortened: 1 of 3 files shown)


src/components/formly.form.ts
```diff
@@ -101,11 +101,29 @@ export class FormlyForm implements OnInit, OnChanges {
...
+        validators.push(validatorFn(this.form, field.validators[validator], validator));
+      }
+    }
     if (Array.isArray(field.validation)) {
-      let validators = [];
       field.validation.map((validate) => {
         validators.push(this.formlyConfig.getValidator(validate).validation);
       });
+    }
+    if (field.validators || Array.isArray(field.validation)) {
```
aitboudad: testing `validators`  here is enough

src/components/formly.form.ts
```diff
@@ -101,11 +101,29 @@ export class FormlyForm implements OnInit, OnChanges {
...
+    let validators = [];
+    if (field.validators) {
+      let validatorFn = function(form, fn, validator) {
+        return (changes) => {
+          if (fn(changes, form)) {
+            return null;
+          } else {
+            let result = {};
+            result[validator] = true;
+            return result;
```
aitboudad: `return { [validator]: true }`

demo/hello.ts
```diff
@@ -179,6 +179,14 @@ export class HelloApp {
             label: 'Zip',
             placeholder: '76010',
           },
+          validators: {
+            zipCode: (changes) => {
```
aitboudad: use `control` as name instead of changes
```ts
zipCode: (control: FormControl) => {
```

src/components/formly.form.ts
mohammedzamakhan: The user might not have defined field.validators, but might have field.validation which is an array. So we still require this.

aitboudad: I mean `validators` and not `field.validators`
```ts
if (validators.length > 0) {
```
see #167

Issue #171, opened 2016-10-26
Adding more templateOptions


mohammedzamakhan:
Adding more templateOptions like 
- [x] Required (#174)
- [x] Max (#174)
- [x] Min (#174)
- [x] Maxlength (#174)
- [x] Minlength (#174)
- [x] Tabindex
- [x] Pattern (#174)
- [x] Step

PR #179, opened 2016-10-26
feat(TemplateOption) added tabindex, placeholder, step to be handled …

mohammedzamakhan:
…using elementRef

### DIFF
--- a/demo/hello.ts
+++ b/demo/hello.ts
@@ -3,7 +3,7 @@ import 'core-js/es6';
 import 'core-js/es7/reflect';
 import 'zone.js/dist/zone';
 
-import { NgModule, Component, ViewChild, ViewContainerRef } from '@angular/core';
+import { NgModule, Component, ViewChild, ViewContainerRef, enableProdMode } from '@angular/core';
 import { FormsModule, ReactiveFormsModule, Validators, FormBuilder, FormGroup, FormControl } from '@angular/forms';
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { BrowserModule } from '@angular/platform-browser';
@@ -148,8 +148,9 @@ export class HelloApp {
           type: 'input',
           templateOptions: {
             type: 'password',
-            label: 'Password',
+            label: 'Confirm Password',
             placeholder: 'Confirm Password',
+            tabindex: 1,
           },
           validators: {
             validation: ValidationService.confirmPassword(this.form, 'password'),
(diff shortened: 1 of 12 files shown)


mohammedzamakhan:
#171

mohammedzamakhan:
Have to upgrade the repo again as we did not export `FieldType` class 😢

demo/hello.ts
```diff
@@ -385,5 +387,5 @@ export class HelloApp {
 })
 export class FormlyDemoModule {
 }
-
+enableProdMode();
```
aitboudad: is this necessary ?

src/templates/field.type.ts
```diff
@@ -1,8 +1,15 @@
 import { Field } from './field';
-import { OnInit, OnChanges, DoCheck, AfterContentInit, AfterContentChecked, AfterViewInit, OnDestroy, AfterViewChecked } from '@angular/core';
+import { OnInit, OnChanges, DoCheck, AfterContentInit, AfterContentChecked, AfterViewInit, OnDestroy, AfterViewChecked, Renderer, ElementRef } from '@angular/core';
 
 export abstract class FieldType extends Field implements OnInit, OnChanges, DoCheck, AfterContentInit, AfterContentChecked, AfterViewInit, AfterViewChecked, OnDestroy {
-
+  private _renderer: Renderer;
+  private _elementRef: ElementRef;
+  private attrbites = ['placeholder', 'tabindex', 'step'];
+  constructor(renderer: Renderer, elementRef: ElementRef) {
```
aitboudad: I don't prefer this way :), instead use a custom directive like we have been doing for focus `formlyNgFocus`

src/templates/field.type.ts
```diff
@@ -27,6 +34,11 @@ export abstract class FieldType extends Field implements OnInit, OnChanges, DoCh
   }
 
   ngAfterViewInit() {
+    const input = this._elementRef.nativeElement.querySelectorAll('input,select,textarea');
+    this.attrbites.filter(attribute => this.field.templateOptions[attribute] !== '' || this.field.templateOptions[attribute] !== undefined)
+      .map(attribute => {
+        this._renderer.setElementAttribute(input[0], attribute, this.field.templateOptions[attribute]);
```
aitboudad: we should deal the update of `templateOptions`

demo/hello.ts
mohammedzamakhan: That will make sure that our demo will run in prod mode as well, I made some changes which were not running in prod mode, so having this could help us develop for production apps.

Issue #183, opened 2016-10-27
Add an event on form controls ready


Riron:
In order to know when the form controls have been registered, it would be nice to have something like a `(controlsReady)` event. It would be emitted after [registerFormContols](https://github.com/formly-js/ng2-formly/blob/2b200b23421175c30febc2a03404b2baafd6986c/src/components/formly.form.ts#L44) has completed I guess.

This would enable us to observe controls valuesChanges safely.

mohammedzamakhan:
@Riron nice idea, can you make a PR and update the demo with a nice example. Even I was not able to observe the controls, I had to observe the form and then observe the control if available.

aitboudad:
Instead you can use `lifecycle` see https://github.com/formly-js/ng2-formly/pull/151

PR #185, opened 2016-10-27
Add an event on form controls ready

Riron:
See #183


aitboudad:
I'm not sure if it's a good idea!

mohammedzamakhan:
Could there be any other use case for this if, this can already be done in lifecycle?

Riron:
Indeed the lifecycles integration seems to do just fine !

PR #187, opened 2016-10-27
feat(FormlyField) generate unique id's for the form"

mohammedzamakhan:


src/components/formly.form.ts
```diff
@@ -7,34 +7,30 @@ import { FormlyConfig } from '../services/formly.config';
...
-  ngOnInit() {
-    if (!this.formModel) {
-      this.formModel = this.model;
-    }
-  }
-
   ngOnChanges(changes: SimpleChanges) {
     if (changes['fields']) {
       this.registerFormControls(this.fields, this.form, this.model);
+      this.formId = `formly_${this.formlyConfig.formId++}`;
```
aitboudad: It can be moved into `ngOnInit`, generate id on each changes doesn't make sense

src/services/formly.config.ts
```diff
@@ -7,6 +7,7 @@ export const FORMLY_CONFIG_TOKEN = new OpaqueToken('FORMLY_CONFIG_TOKEN');
  */
 @Injectable()
 export class FormlyConfig {
+  public formId = 0;
```
aitboudad: is there any use case here ? as I think it can be moved to `FormlyForm`

src/templates/types/input.ts
```diff
@@ -4,9 +4,12 @@ import { FieldType } from '../field.type';
 @Component({
   selector: 'formly-field-input',
   template: `
-    <input [type]="templateOptions.type" [formControl]="formControl" class="form-control" id="{{key}}"
+    <input [type]="type" [formControl]="formControl" class="form-control" id="{{id}}"
```
aitboudad: I prefer using `[id]="id"` instead of `id="{{id}}"`

src/templates/types/select.ts
```diff
@@ -4,7 +4,7 @@ import { FieldType } from '../field.type';
 @Component({
   selector: 'formly-field-select',
   template: `
-    <select [id]="key" [formControl]="formControl" class="form-control" [formlyAttributes]="templateOptions">
+    <select id="{{id}}" [id]="key" [formControl]="formControl" class="form-control" [formlyAttributes]="templateOptions">
```
aitboudad: duplicate `id` attribute `id="{{id}}" [id]="key"`

src/components/formly.field.ts
```diff
@@ -130,6 +140,7 @@ export class FormlyField implements DoCheck, OnInit {
         form: this.form,
         field: this.field,
         formModel: this.formModel,
+        id: this.id,
```
aitboudad: instead we can set `id` in the field itself `this.field.id = this.id`

src/components/formly.form.ts
```diff
@@ -7,34 +7,30 @@ import { FormlyConfig } from '../services/formly.config';
...
     <ng-content></ng-content>
   `,
 })
-export class FormlyForm implements OnInit, OnChanges {
-  @Input() formModel: any;
+export class FormlyForm implements OnChanges {
   @Input() model: any = {};
   @Input() form: FormGroup = new FormGroup({});
   @Input() fields: FormlyFieldConfig[] = [];
+  public formId;
```
aitboudad: just `formId` no need the `public` keyword here

PR #196, opened 2016-10-29
refactor(Validation): rename FromlyMessage to FormlyValidationMessage

aitboudad:


mohammedzamakhan:
I was wondering if FormlyValidationMessage should be a in components folder or templates folder. I think it should be in templates because other templates might not need a div, they might want to use a span with some classes or something.

aitboudad:
let's keep it in components folder, the extra div is removed now
```ts
template: `{{errorMessage}}`,
```

mohammedzamakhan:
Also think about this, some projects might want to show all the messages to the user, right now our component only shows one. This is definitely a part of template folder since even angular-formly had it in templates.

aitboudad:
make sense now :+1: , done!

PR #198, opened 2016-10-29
refactor(bootstrap): move bootstrap templates to `ui-bootstrap` folder

aitboudad:
selection_036


mohammedzamakhan:
I think formly.validation-message.ts should not be in core templates, it should be in ui-bootstrap folder, because its a project specific implementation, and has nothing to do with core functionality.

mohammedzamakhan:
Except that that looks great to me 👍

Issue #204, opened 2016-10-29
field and fieldGroup wrapper


mohammedzamakhan:
field wrapper: http://angular-formly.com/#/example/other/async-select-options-with-controller
field group wrapper: http://angular-formly.com/#/example/other/nested-formly-forms

PR #205, opened 2016-10-31
feat(FieldWrapper) FieldWrapper

mohammedzamakhan:
#204

### DIFF
--- a/src/core/components/formly.field.ts
+++ b/src/core/components/formly.field.ts
@@ -16,16 +16,6 @@ import 'rxjs/add/operator/debounceTime';
   template: `
     <template #fieldComponent></template>
     <div *ngIf="field.template && !field.fieldGroup" [innerHtml]="field.template"></div>
-
-    <formly-field *ngFor="let f of field.fieldGroup; let i = index"
-      [hide]="f.hideExpression"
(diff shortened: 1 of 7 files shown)


src/core/components/formly.field.ts
```diff
@@ -113,6 +103,8 @@ export class FormlyField implements DoCheck, OnInit {
       });
 
       this.formlyPubSub.setEmitter(this.field.key, update);
+    } else if (this.field.fieldGroup) {
```
aitboudad: we can group it with the `this.field.fieldArray` test
```ts
else if (this.field.fieldGroup || this.field.fieldArray)
```

src/core/components/formly.field.ts
```diff
@@ -121,7 +113,9 @@ export class FormlyField implements DoCheck, OnInit {
   private createFieldComponent(): ComponentRef<Field> {
     // TODO support this.field.hideExpression as a callback/observable
     this.hide = this.field.hideExpression ? true : false;
-
+    if (this.field.fieldGroup) {
+      this.field.type = 'formly-group';
```
aitboudad: set type only when it's empty

PR #212, opened 2016-11-01
Prevent collisions between multiple radio groups and multicheckbox groups

danielcrisp:
Fixes https://github.com/formly-js/ng2-formly/issues/210

Previously if there were two or more radio or checkbox lists, with a different `key`, but with the same values then changing one list would affect the other list

### DIFF
--- a/demo/hello.ts
+++ b/demo/hello.ts
@@ -88,7 +88,7 @@ export class HelloApp {
 
     let userFields: Array<FormlyFieldConfig> = [{
         type: 'radio',
-        key: 'title',
+        key: 'title1',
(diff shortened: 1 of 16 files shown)


mohammedzamakhan:
Please resolve the conflict and push again.

danielcrisp:
You got it :-)

danielcrisp:
Is it possible to add a new tag so I can get this change with yarn / npm please?

mohammedzamakhan:
Sure 👍

mohammedzamakhan:
Done

PR #216, opened 2016-11-01
feat(FormlyConfig) setType can accept array of options

mohammedzamakhan:


src/core/services/formly.config.ts
```diff
@@ -31,17 +32,23 @@ export class FormlyConfig {
...
-    this.types[options.name].extends = options.extends;
-    if (options.wrappers) {
-      options.wrappers.map((wrapper) => {
-        this.setTypeWrapper(options.name, wrapper);
+  setType(options: TypeOption | TypeOption[]) {
+    if (Array.isArray(options)) {
+      options.map(option => {
+        this.setType(option);
       });
+    } else if (this.formlyUtils.isObject(options)) {
```
aitboudad: this test is useless here, if `options` is not an array then it's an instance of TypeOption
```ts
   ...
} else {
   if (!this.types[options.name]) {
   ...
}
```

aitboudad: fixed in #217

PR #217, opened 2016-11-01
refactor(FormlyConfig): remove useless `if` statement

aitboudad:


src/core/services/formly.config.ts
```diff
@@ -37,7 +36,7 @@ export class FormlyConfig {
       options.map(option => {
         this.setType(option);
       });
-    } else if (this.formlyUtils.isObject(options)) {
+    } else {
```
aitboudad: well I'm not sure if it is the right way!, typescript is used to solve all issues related to the typing and I think angular itself doesn't care about that

PR #224, opened 2016-11-02
feat(Events) add events to TemplateOptions

### LINKED ISSUE #147: Handling onKeypress, onEnter, onLeave, onClick, onFocus, onBlur, onChange etc.
Lets discuss how can we deal with this. Angular 1 formly had this to tackle it, but we don't have anything in place yet.

``` js
     {
        key: 'text1',
        type: 'input',
        templateOptions: {
          onKeypress: function($viewValue, $modelValue, scope) {
            console.log('The key was pressed!', $viewValue);
          }
        }
      }
```

mohammedzamakhan:
closes #147

### DIFF
--- a/demo/hello.ts
+++ b/demo/hello.ts
@@ -121,6 +121,9 @@ export class HelloApp {
             type: 'password',
             label: 'Password',
             placeholder: 'Password',
+            keyup: (field, formControl: FormControl) => {
+              console.log(formControl.valid ? 'Valid' : 'Invalid');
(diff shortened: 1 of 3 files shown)


jrgleason:
Is there a reason we used `changed` instead of `onChange`?

mohammedzamakhan:
@jrgleason because `change` is used to attach events in plain JavaScript, is their specific use case for `onChange`

PR #228, opened 2016-11-06
feat(defaultOption): implements defaultOption

### LINKED ISSUE #200: defaultOption for a type
http://angular-formly.com/#/example/custom-types/default-options

divyakumarjain:
implements defaultOption which helps in defining reusable types and keeping configuration short

Closes #200

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 npm-debug.log
 yarn-error.log
 coverage
+typings
--- a/demo/hello.ts
+++ b/demo/hello.ts
@@ -146,19 +146,9 @@ export class HelloApp {
           template: '<br/><hr/>',
         }, {
           className: 'col-md-4',
-          key: 'select',
           type: 'select',
-          templateOptions: {
-            options: [{
-              label: 'Male',
-              value: 'male',
-            }, {
-              label: 'Female',
-              value: 'female',
-            }],
-            label: 'Gender',
-            placeholder: 'Select Gender',
-          },
+          optionsType: ['gender'],
+          key: 'select',
         }],
       }, {
         className: 'section-label',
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ node_modules
 npm-debug.log
 yarn-error.log
 coverage
+typings
```
aitboudad: We don't use typings anymore, instead can you remove typings folder in your local

package.json
```diff
@@ -22,9 +22,9 @@
     "build:lib": "npm run ngc -- --project tsconfig.build.json",
     "prebuild:bundles": "shx rm -rf ./bundles",
     "build:bundles": "npm run build:bundles:compile && npm run build:bundles:minify",
-    "build:bundles:compile": "node ./node_modules/.bin/webpack --config .config/webpack.build.js",
+    "build:bundles:compile": "webpack --config .config/webpack.build.js",
     "build:bundles:minify": "uglifyjs bundles/ng2-formly.umd.js --output bundles/ng2-formly.umd.min.js --in-source-map bundles/ng2-formly.umd.js.map --source-map bundles/ng2-formly.umd.min.js.map",
-    "demo": "NODE_ENV=dev node node_modules/.bin/webpack-dashboard --title ng2-formly-demo -- node_modules/.bin/webpack-dev-server --quiet --port 9000 --content-base demo --config .config/webpack.demo.js --open",
+    "demo": "webpack-dashboard --title ng2-formly-demo -- node_modules/.bin/webpack-dev-server --quiet --port 9000 --content-base demo --config .config/webpack.demo.js --open",
```
aitboudad: This will require to install webpack-dashboard globally , I prefer the old one

src/core/components/formly.field.ts
```diff
@@ -57,9 +57,23 @@ export class FormlyField implements DoCheck, OnInit {
   }
 
   ngOnInit() {
+    this.initDefaultOptions();
```
aitboudad: I would prefer to move this into formly-form component

demo/hello.ts
```diff
@@ -461,6 +451,22 @@ export class HelloApp {
         { name: 'formly-wrapper-horizontal', component: FormlyWrapperHorizontalLabel, types: ['horizontalInput'] },
         { name: 'panel', component: FormlyPanelWrapper },
       ],
+      defaultOptions: [
+        {
```
mohammedzamakhan: defaultOptions is not meant to be set separately, its set as a part of the setType.

package.json
divyakumarjain: Reverting it but i was not able to run old one one windows machine.

demo/hello.ts
divyakumarjain: @mohammedzamakhan after looking at PR #232 seems like the direction in this PR is different then intended. Let me know if you want to me to close this PR and you will continue with #232 .

package.json
divyakumarjain: raised new issue https://github.com/formly-js/ng2-formly/issues/233 for this

PR #232, opened 2016-11-08
defaultOptions

mohammedzamakhan:


mohammedzamakhan:
@divyakumarjain This is how I wanted to have defaultOptions (similar to how its in angular-formly [ng1]).

I am seeing some issues in this, but let me know if you want to use this for the implementation. I can work on something else.

divyakumarjain:
Seems that our comment crossed eachother, I just replied the other thread :). Let me see if i can take this forward. If i am not able to finish then I will be on vacation until next weekend.

mohammedzamakhan:
I used reverseDeepMerge from our father repository (angular-formly) by @kentcdodds 😺 

https://github.com/formly-js/angular-formly/blob/master/src/other/utils.js#L43

divyakumarjain:
what is the issue which you are seeing in this?

mohammedzamakhan:
The validation wasn't being applied from the type `dateFormat` to Investment Date, sometimes it does apply the validation and sometimes it doesn't and also the toggleVal field has `isLarge: false` which should not be overridden by its type which has `isLarge: true`

PR #281, opened 2016-11-23
feat(HideExpression) add consistent context to function expression

franzeal:
Feature

When hideExpression is a function, `this` and `args` of the function are inconsistent with the context when the expression is a string and does not seem to provide useful access to the field.

When hideExpression is a function it is called with the same `this` and `args` as it would be if it were a string.

**Other information**:
This may not be necessary when a hideExpression function is declared in the templateOptions for a field; however, this is intended to address the behavior observed when the hideExpression is declared with a function within the defaultOptions of the formly config.

### DIFF
--- a/src/core/services/formly.field.delegates.ts
+++ b/src/core/services/formly.field.delegates.ts
@@ -7,7 +7,7 @@ export class FormlyFieldVisibilityDelegate {
   eval(expression: string | Function | boolean): boolean {
     // TODO support this.formlyCommon.field.hideExpression as a observable
     if (expression instanceof Function) {
-      return expression();
+      return expression.apply(this.formlyCommon, [this.formlyCommon.model, this.formlyCommon.options.formState]);
     } else if (typeof expression === 'string') {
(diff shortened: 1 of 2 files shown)


src/core/services/formly.field.delegates.ts
```diff
@@ -7,7 +7,7 @@ export class FormlyFieldVisibilityDelegate {
   eval(expression: string | Function | boolean): boolean {
     // TODO support this.formlyCommon.field.hideExpression as a observable
     if (expression instanceof Function) {
-      return expression();
+      return expression.apply(this.formlyCommon, [this.formlyCommon.model, this.formlyCommon.options.formState]);
```
aitboudad: please update [FormlyFieldConfig](https://github.com/formly-js/ng2-formly/blob/master/src/core/components/formly.field.config.ts#L14)
```ts
hideExpression?: boolean | string | ((model, formState) => boolean);
```

mohammedzamakhan:
How is this feature related to validation?

franzeal:
@mohammedzamakhan Sorry about the incorrect naming.

aitboudad:
Thank you @franzeal

Issue #230, opened 2016-11-07
Model binding to a formly-field in FormlyForm will not work with a nested property


Krustie101:
In the template of  FormlyForm the model of the formly-field is bound via the expression

`[model]="field.key?model[field.key]:model"`

If field.key is a nested property (i.e. contains dots), model[field.key] will not be evaluated properly. I think a custom pipe could be used to handle nested properties (applied to model and field.key as a parameter).

aitboudad:
define custom method is enough, can you work on it ?

Krustie101:
@aitboudad 
No problem, I will take it up.

mohammedzamakhan:
Any updates on this @Krustie101 ?

Krustie101:
@mohammedzamakhan 

See pull request #293

PR #293, opened 2016-11-26
#230 added methods to extract nested properties & create empty objects

Krustie101:
Bug fix. In formly form, the fieldModel method assumes that the key is not nested. This bugfix also deals with the case that the key is a nested property (including support for indices, if you really want to use this). The main code is located in the utils class.

Tests have been added for the extra code in the utils class.

See #230 

**Other information**:


mohammedzamakhan:
Removed all the linting errors in a PR to your repo https://github.com/Krustie101/ng2-formly/pull/1

Please run the demo app, because I don't see it running with this update.

Krustie101:
@mohammedzamakhan 
When I try to run the demo task on either my fork or the original ng2-formly code, I get the error "Uncaught Error: Can't resolve all parameters for FormlyValidationMessages:".

After debugging,  the root cause seems to be the fact that webpack loads 2 instances of the angular module. When angular tries to resolve the dependencies for FormlyValidationMessages, the instanceof `@Inject` check fails because there are 2 instances of this decorator.

So I wonder how you manage to run this demo task?

mohammedzamakhan:
I don't get that error neither do @aitboudad. I wonder what could cause this error. I was able to run your fork as well

franzeal:
@Krustie101 I see the same in my environment.

mohammedzamakhan:
@divyakumarjain would know about it since he works on windows. He had made these changes in the build, try using it
https://github.com/mohammedzamakhan/ng2-formly/pull/2/files

franzeal:
@mohammedzamakhan I was able to get it running using those changes to the files you referenced.  I did end up using the current package.json's build:bundles:compile task definition as well as adding `<script src="/build/vendor.js"></script>` to demo/index.html right before the demo script.

That got the demo application running to the point where I am getting an error stemming from a validation expression in the demo:  "Cannot set property 'custom' of undefined"

mohammedzamakhan:
I just made the changes in my Windows VM, and I didn't get any such error.

mohammedzamakhan:
@Krustie101 update your fork with latest, the demo application should run on windows too now 👍

franzeal:
@mohammedzamakhan can confirm it runs in Windows now without error.  Good work and thanks!

Krustie101:
@mohammedzamakhan  Thank you, it also works on my end.  I will now have a look at the other error.

mohammedzamakhan:
@Krustie101 it worked for me when I did

```js
fieldModel(field: FormlyFieldConfig) {
     if (field.key && (field.fieldGroup || field.fieldArray)) {
       return getFieldModel(this.model, field, true);
     }
     return this.model;
}
```

Krustie101:
@mohammedzamakhan That is correct. I also noticed my error, arising from the fact that I did not pay attention that you had changed the meaning of model.

mohammedzamakhan:
LGTM 👍 

@aitboudad can you have a look at this PR and approve it?

aitboudad:
@mohammedzamakhan I can't right now :( you can merge it if it looks good, 
I'll give my feedback, once I get some time!

mohammedzamakhan:
@Krustie101 can you also look into using this pattern for [assignModelValue in utils.ts#L67](https://github.com/Krustie101/ng2-formly/blob/105197bfe607e06aca95af85312564e7f37b9675/src/core/utils.ts#L67) and [getValueForKey  in utils.ts#L83](https://github.com/Krustie101/ng2-formly/blob/105197bfe607e06aca95af85312564e7f37b9675/src/core/utils.ts#L83)

Krustie101:
@mohammedzamakhan  I will do, but I am tied up tomorrow. I will look at in on Thursday.

Krustie101:
The problem lies in the use of formControlName in the FormlyFieldRadio component, this cannot be a nested path. Like in the case of an input the formControlName should be removed and [formControl]="formControl"  should be used instead.

I also notice on line 56 of the FormBuilder (master branch) the following check: 

` if (!form.get(field.key)) {`

Should this not be 

` if (!form.get(field.key)) {`

Issue #308, opened 2016-12-05
Not getting Typescript and Webpack config when using github and npm


jrgleason:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[X] feature request
[ ] support request
```

**Current behavior**

1. Add `"ng2-formly": "git://github.com/formly-js/ng2-formly.git"` to package.json
2. Go into node_modules/ng2-formly 
3. Run `npm install`
4. Notice it complains about the Typescript config.

**Workaround**

Copy files from a typical clone of the git repo

**Expected behavior**
I should get a valid build without work around

I tried this on OSX and Windows.

jrgleason:
Looks like this is the culprit....

https://github.com/formly-js/ng2-formly/blob/master/.npmignore

mohammedzamakhan:
Please create a PR for the change 😄

jrgleason:
Will do, are we sure we want to allow these files to be downloaded over npm? Do we just need the build typescript config or the regular one as well? What Webpack configs should be included? All?

jrgleason:
There is your pull request... It seems cheap to become a contributing coder on something this menial.

mohammedzamakhan:
@jrgleason yayyy!!! Thanks 👍

jrgleason:
Still some issues unfortunately. I tried to only add a couple files and I was assuming some things would happen. This was not the case, however, I am curious why spec files are being used if we are not supposed to be callings tests when running build. 

So the real question is, why would I get...

```
ERROR in [at-loader] src/core/utils.spec.ts:5:3 
    Cannot find name 'describe'.
```

When running npm run build? shouldn't I need to run npm run test?

mohammedzamakhan:
Make sure that you make changes to your repo and make changes to it and instead of 

```js
"ng2-formly": "git://github.com/formly-js/ng2-formly.git"
```

use 

```js
"ng2-formly": "git://github.com/jrgleason/ng2-formly.git"
```

and test it out in your next PR 😄

PR #309, opened 2016-12-05
These need to be added if we want to build locally issue #308

jrgleason:

Allows us to build from the npm artifact

You need to copy over the webpack and typescript configuration

The configuration is downloaded on npm install.

- [?] The commit messages follow our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format

**Other information**:

I think that we could use a lookBack regex on the .config folder but this is my lazy version.

https://github.com/formly-js/ng2-formly/issues/308

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,6 @@
 .idea
-.config
+.config\karma-shim.js
+.config\webpack.test.js
 demo
 node_modules
 coverage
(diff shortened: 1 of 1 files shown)


jrgleason:
Looks like I am still missing some files, demo seems required looking into others.

jrgleason:
NM looks like it is working can you confirm? Remember tests and such shouldn't run just build.

PR #324, opened 2016-12-09
feat(field): add hide option

aitboudad:

http://docs.angular-formly.com/docs/field-configuration-object#hide-boolean


mohammedzamakhan:
please add an example in the demo app too. Thanks

aitboudad:
done

Issue #333, opened 2016-12-10
Arrays in Property Keys


franzeal:
**I'm submitting a ...**
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request
```

**Current behavior**
When a field's key contains an array indexer, formly does not treat that part of the key as an array.

**Expected behavior**
When a field's key contains an array indexer, formly should treat that part of the key as an array.

**Minimal reproduction of the problem with instructions**
https://plnkr.co/edit/AnIVMgao1dvjHsSkTDHh?p=preview

**What is the motivation / use case for changing the behavior?**
I have an existing angular-formly form that uses this capability that I'm porting.

mohammedzamakhan:
For now you can do this like https://plnkr.co/edit/XhsLoifpOR29Wyr0uY0V?p=preview

franzeal:
Sorry, I know I've seen that suggested before.  Workaround works fine, much appreciated!

mohammedzamakhan:
@Krustie101 is expert in solving this issue 😅 

#293 PR had a solution to similar problem, but applied to different scenario.

PR #343, opened 2016-12-15
fix(Array) Array in property access

mohammedzamakhan:
#333

### DIFF
--- a/demo/hello.ts
+++ b/demo/hello.ts
@@ -360,12 +360,20 @@ export class HelloApp {
         },
       },
       {
-        key: 'nested.arrays.0',
+        key: 'nested.arrays.0.item',
+        type: 'input',
+        templateOptions: {
+          label: 'Array with dots',
+        },
+        defaultValue: 'Default Value A',
+        parsers: [(value) => (value || '').toUpperCase()],
+      }, {
+        key: 'nested.arrays[1].item',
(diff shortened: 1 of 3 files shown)


mohammedzamakhan:
@Krustie101

demo/hello.ts
```diff
@@ -360,12 +360,20 @@ export class HelloApp {
...
-        key: 'nested.arrays.0',
+        key: 'nested.arrays.0.item',
+        type: 'input',
+        templateOptions: {
+          label: 'Array with dots',
+        },
+        defaultValue: 'Default Value A',
+        parsers: [(value) => (value || '').toUpperCase()],
+      }, {
+        key: 'nested.arrays[1].item',
```
aitboudad: I'm not in favor of this, isn't `nested.arrays.1.item` enough ?
also this make us consistent with angular form

aitboudad: the easy way to solve this without lot of changes is replacing `[` and `]` by `.`
`nested.arrays[1].item` -> `nested.arrays.1.item`

mohammedzamakhan: I just wanted to keep it consistant with angular-formly, also most developers are used to use [] for arrays

aitboudad: make sense :+1:

PR #366, opened 2017-01-26
fix attr multiple for tag select

maranqz:
fix https://github.com/formly-js/ng-formly/issues/365

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ng-formly/366)

### DIFF
--- a/src/core/components/formly.attributes.ts
+++ b/src/core/components/formly.attributes.ts
@@ -9,7 +9,7 @@ import { FormlyFieldConfig } from './formly.field.config';
 export class FormlyAttributes implements OnInit, OnChanges {
   @Input('formlyAttributes') field: FormlyFieldConfig;
   @Input() formControl;
-  private attributes = ['id', 'name', 'placeholder', 'tabindex', 'step', 'aria-describedby'];
+  private attributes = ['id', 'name', 'placeholder', 'tabindex', 'step', 'aria-describedby', 'multiple'];
   private statements = ['change', 'keydown', 'keyup', 'keypress', 'click', 'focus', 'blur'];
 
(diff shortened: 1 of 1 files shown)


mohammedzamakhan:
Can you put an example in the demo, because I don't think just allowing multiple attribute will make it work

maranqz:
Yes. It's mad idea cause we have only one option value.
```value = "a"; /* but no */ value = ["a", "b"];```
Unforchanatly [Dynamically created add directive for component](https://github.com/angular/angular/issues/7329)
It is two way resolve this problem.
Using own "ControlValueAccessor " with twice([multiple ](https://github.com/angular/angular/blob/46023e47925c74028ab2c07eeaab96c9f89443c3/modules/%40angular/forms/src/directives/select_control_value_accessor.ts), [single](https://github.com/angular/angular/blob/821b8f09d6054261c737f30ce500d912d7d8017e/modules/%40angular/forms/src/directives/select_multiple_control_value_accessor.ts)) type select or using <template [ngIf]>.
What's better?
I realised second [variant](https://plnkr.co/edit/viRK9wswqxu3ClveOLPe?p=preview). 
I can't using ngTemplateOutlet cause defaultValue don't set.
It's normal if a change my pull then will you accept it?

franzeal:
This seems pretty heavy handed.  Wouldn't it be more appropriate for the select field type to specifically support this attribute or to just create your own multi-select type that allows for this?

I've created a multi-select field type (I've opted to use all my own field types for our project) rather than try to make the select field type versatile enough to handle both use cases.

aitboudad:
I'll provide a better way, closing now!

Issue #408, opened 2017-04-27
reverseDeepMerge throws an exception when dealing with readonly property 


ericof:
**I'm submitting a ...**
```
[X] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request
```

**Current behavior**
Using the following model

```
class FooBar {

  foo = 'foo';
  bar = 'bar';

  get foobar(): string {
    return '';
  }
}
```

reverseDeepMerge will throw a TypeError exception when setting the value of foobar.

**Expected behavior**
reverseDeepMerge will not throw the exception and will proceed with the merge of src and dest.

**Minimal reproduction of the problem with instructions**
A form with a model that contains either readonly property or a getter without a setter and existing data.

**What is the motivation / use case for changing the behavior?**
It is a common use case to have models implementing readonly properties.

**Please tell us about your environment:**
* **Operating System:**: Linux, osX

* **Angular version:**: Angular 2.x, Angular 4.x

* **Browser:**: all
 
* **Language:**: TypeScript 2.0.10, TypeScript 2.2.0

PR #409, opened 2017-04-27
fix(utils): deal with readonly properties on reverseDeepMerge (#408)

ericof:
fix for issue #408 

reverseDeepMerge throws an exception when a property on dest is readonly, empty or null.

when the model contains a readonly property and the default dest is empty, reverseDeepMerge will
catch the exception and proceed with the merge

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ng-formly/409)

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -113,7 +113,13 @@ export function reverseDeepMerge(dest, source = undefined) {
         if (isFunction(src[srcArg])) {
           dest[srcArg] = src[srcArg];
         } else {
-          dest[srcArg] = clone(src[srcArg]);
+          try {
+            dest[srcArg] = clone(src[srcArg]);
+          }  catch (e) {
+            if (e instanceof TypeError) {
+              // Setting a readonly attribute
+            }
+          }
         }
       } else if (objAndSameType(dest[srcArg], src[srcArg])) {
         reverseDeepMerge(dest[srcArg], src[srcArg]);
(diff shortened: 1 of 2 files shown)


src/core/utils.ts
```diff
@@ -113,7 +113,13 @@ export function reverseDeepMerge(dest, source = undefined) {
         if (isFunction(src[srcArg])) {
           dest[srcArg] = src[srcArg];
         } else {
-          dest[srcArg] = clone(src[srcArg]);
+          try {
```
aitboudad: I'm not sure if `try-catch` is the better approach here, can we check the type of prop instead ?

ericof: I tried to find a sane/simple way of checking if the property was readonly but found none, that is why I decided for the try catch.

aitboudad: not sure if this is the right fix as it hide a message that can be useful in some cases

aitboudad: another approach that we can consider is avoid setting the model value if the field config is marked as `readonly`

ericof: The problem that led to this pr happens with a property that is not even used on the form itself. (and some of my models are a bit complex and adding hidden fields just to set readonly  would be counterproductive).

ericof: Also, I considered leaving a console.log inside the catch but it seemed too verbose.

aitboudad: @ericof can you provide me a reproducing example with a property which is not used on the form I'm pretty sure we can find a better way

PR #533, opened 2017-10-20
fix(FormlyFormBuilder): address null reference when `validation.show`…

franzeal:
… is true

Addresses a null reference introduced by #417 that can occur when `validation.show` is true.

Bug fix

Attempts to call `markAsTouched` on a null object reference.

Calls `markAsTouched` on the existing reference.

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ng-formly/533)


aitboudad:
published `1.0.0-rc.14`

PR #586, opened 2017-12-06
fix(core): remove usage of setTimeout.

aitboudad:
**What kind of change does this PR introduce? Bug fix**

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ngx-formly/586)


mohammedzamakhan:
setTineout looks ugly? Do you have any other solution in mind too?

aitboudad:
@mohammedzamakhan yep there is another better approach but it's require more work with a potential breaking change, let me think on it!

aitboudad:
setTimeout is removed now, I hope this will not produce some unexpected behavior ;)

PR #606, opened 2017-12-21
fix(tsconfig.json): Angular modules added to "paths"

### LINKED ISSUE #605: Error during webpack compilation of the demo app
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request
```

**Current behavior**

Error during "yarn demo":

> ERROR in Error: Error encountered resolving symbol values statically. Calling function 'InjectionToken', function calls are not supported.

**Expected behavior**

- Clone the ngx-formly repo
- Run "yarn"
- Run "yarn demo"

* **Angular version:** 2.0.X

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =   v9.3.0

AlexTalcura:
Angular modules added to "paths" section of tsconfig.json in order to avoid "Error encountered
resolving symbol values statically. Calling function 'InjectionToken', function calls are not
supported." error

fix #605

**Other information**:

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ngx-formly/606)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,6 +14,7 @@
     "lib": ["dom", "es6"],
     "importHelpers": true,
     "paths": {
+      "@angular/*": ["../node_modules/@angular/*"],
       "@ngx-formly/core": ["./core/index"],
       "@ngx-formly/bootstrap": ["./ui-bootstrap/index"],
       "@ngx-formly/material": ["./ui-material/index"]
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,7 +14,10 @@
     "lib": ["dom", "es6"],
     "importHelpers": true,
     "paths": {
+      // fix for #605
+      // TODO: Remove it after upgrading Angular to v.5      
       "@angular/*": ["../node_modules/@angular/*"],
+      
       "@ngx-formly/core": ["./core/index"],
       "@ngx-formly/bootstrap": ["./ui-bootstrap/index"],
       "@ngx-formly/material": ["./ui-material/index"]


aitboudad:
I'm not agree with the current changes, probably there is something wrong in your workspace, can you try to remove `node_modules` and install packages again + ensure you use `yarn` and not `npm`

AlexTalcura:
This was a clean install (i.e. no node_modules) folder at all. I used "yarn" and then "yarn demo". Didn't use npm commands directly at all.

AlexTalcura:
Just to make clear, I don't have such line in other angular apps (i.e. work projects) but I can't build this repo without this line. Tried even to build it with this line and then remove the line - new "yarn demo" will fail (see below). Maybe you can try to do a clean build yourself and see if it works in your environment?

Few notes:
Win 10
Node v.9.3.0
NPM v.5.5.1
Yarn v.1.3.2

image

aitboudad:
@AlexTalcura I'm not able to reproduce it, can you check if the global `@angular/cli` version is the latest one?

AlexTalcura:
Looks like it is 1.6.2

image

aitboudad:
can you try to run demo with:
```batch
node_modules/.bin/ng serve --port 4100 --open
```

AlexTalcura:
Same result. I even did the following:
- deleted node_modules in ngx-formly folder
- then "npm cache clean --force"
- then "yarn"
- and then "yarn demo" and "node_modules\.bin\ng serve --port 4100 --open" in ngx-formly..

Nothing works. I have a question for you: can you please try to build ngx-formly with the line from this commit? Does it work for you? If it still works for you then maybe it makes sense to have in the config in order to help people like me? I do understand that not many people will actually build ngx-formly because most of them will use npm packages, but maybe it worth to keep it?

aitboudad:
I'm trying to find the root cause, probably related to windows but not sure, I would ask you one last things update all angular packages into `v5.0` an try again if it's working I can temporary accept this PR

aitboudad:
Please note that it's an angular issue and it should be reported in order to find the root cause.

AlexTalcura:
Looks like it works with angular 5 in both cases (**with** and **without** this line in the tsconfig)! Here is my package.json for v.5:

[package.txt](https://github.com/formly-js/ngx-formly/files/1580739/package.txt)


> Please note that it's an angular issue and it should be reported in order to find the root cause.
I completely agree with you! However, we need to create some simple app with this issue in order to submit it to them. In fact, this can be related not just to angular but npm as well. And to be honest, I don't think that it worth to spend so much time on it. Especially considering that it is the first time when I encountered such behavior and looks like it is fixed in v.5.

Anyway, it is time for me to finally get to kendo ui implementation for formly : )

tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "lib": ["dom", "es6"],
     "importHelpers": true,
     "paths": {
+      "@angular/*": ["../node_modules/@angular/*"],
```
aitboudad: add comment here that mention the issue number + it should be removed when we use angular 5

AlexTalcura:
``` 
   "paths": {
      // fix for #605
      // TODO: Remove it after upgrading Angular to v.5
      "@angular/*": ["../node_modules/@angular/*"],

      "@ngx-formly/core": ["./core/index"],
      "@ngx-formly/bootstrap": ["./ui-bootstrap/index"],
      "@ngx-formly/material": ["./ui-material/index"]
    },
```

Something like this?

aitboudad: yep

PR #688, opened 2018-02-06
Update formly.field.config.ts

mattmutt:
Can you please verify that consumer projects that have `noImplicitAny` as `true` in their `tsconfig.json` can use this module? I get a few Typescript errors ( code TS7006 ) now as soon as I import the modules from ngxFormly

**Other information**:

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ngx-formly/688)

### DIFF
--- a/src/core/src/components/formly.field.config.ts
+++ b/src/core/src/components/formly.field.config.ts
@@ -32,7 +32,7 @@ export interface FormlyFieldConfig {
    */
   validation?: {
     messages?: {
-      [messageProperties: string]: string | ((error, field: FormlyFieldConfig) => string);
+      [messageProperties: string]: string | ((error: any, field: FormlyFieldConfig) => string);
     };
     show?: boolean;
     [additionalProperties: string]: any;


src/core/src/components/formly.field.config.ts
```diff
@@ -32,7 +32,7 @@ export interface FormlyFieldConfig {
    */
   validation?: {
     messages?: {
-      [messageProperties: string]: string | ((error, field: FormlyFieldConfig) => string);
+      [messageProperties: string]: string | ((error: any, field: FormlyFieldConfig) => string);
```
aitboudad: if we are going on that way, it would be better to use `noImplicitAny` to make sure all arguments are annotated

IustiNN:
There is also this file that needs the same changes: 
node_modules/@ngx-formly/core/src/services/formly.config.d.ts

`node_modules/@ngx-formly/core/src/services/formly.config.d.ts(20,36): error TS7006: Parameter 'error' implicitly has an 'any' type.
node_modules/@ngx-formly/core/src/services/formly.config.d.ts(40,59): error TS7006: Parameter 'error' implicitly has an 'any' type.
node_modules/@ngx-formly/core/src/services/formly.config.d.ts(63,25): error TS7006: Parameter 'error' implicitly has an 'any' type.`

aitboudad:
ok I'll take care of this since it require more work, thanks for the initial work.

PR #705, opened 2018-02-10
perf: avoid using expressionProperties for Cascaded Select example.

aitboudad:
**What kind of change does this PR introduce? docs update**

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ngx-formly/705)

### DIFF
--- a/demo/src/app/examples/other/cascaded-select/app.component.ts
+++ b/demo/src/app/examples/other/cascaded-select/app.component.ts
@@ -1,12 +1,16 @@
-import { Component } from '@angular/core';
+import { Component, OnDestroy } from '@angular/core';
 import { FormGroup } from '@angular/forms';
 import { FormlyFormOptions, FormlyFieldConfig } from '@ngx-formly/core';
+import { Subject } from 'rxjs/Subject';
+import { takeUntil } from 'rxjs/operators/takeUntil';
+import { tap } from 'rxjs/operators/tap';
 
 @Component({
   selector: 'formly-app-example',
   templateUrl: './app.component.html',
 })
-export class AppComponent {
+export class AppComponent implements OnDestroy {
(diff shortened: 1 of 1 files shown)


demo/src/app/examples/other/cascaded-select/app.component.ts
```diff
@@ -89,4 +91,8 @@ export class AppComponent {
   submit() {
     alert(JSON.stringify(this.model));
   }
+
+  ngOnDestroy(): void {
+    this.onDestroy$.complete();
```
juristr: 🤔 I think this won't be enough. You should rather first `this.onDestroy$.next()`, otherwise the observable stream won't be terminated if I'm not wrong.

See: https://gist.github.com/btroncone/fe9d9aaf457f2ebd72c4624e34d664f8

aitboudad: I doubt it, I think `takeUntil` should take account of `complete` too, @benlesh ?

juristr: I had a similar scenario when switching between my Angular application routes. And by just doing the `.complete()` thing, http requests triggered via observable streams didn't get cancelled. Instead doing a `.next()` worked as expected.

@btroncone can shed some light on this for sure?

aitboudad: just tested on an example and seems you were right, I'm going to fix it. thanks!
I often used `complete` without noticing it :sweat_smile:

btroncone:
This is correct, takeUntil triggers off a value being emitted by the inner observable.

This is the pattern I prefer: https://github.com/angular/material2/blob/fa4ddd0a13461b2f846e114fd09f8f4e21b814b1/src/lib/input/autosize.ts#L106

PR #736, opened 2018-02-20
feat(core): allow rendering extra attributes

aitboudad:

There is no way to allow rendering extra html attribures

Introduce `attributes` options which allow users to specify which option in `templateOption` can be used as an attribute.

for example for range type we can use:
```ts
{
  key: 'age',
 type: 'input',
  templateOptions: {
      attributes: {
        min: 5,
        max: 10,
      },
  }
}
```

we can also set those attributes globally in formly declaration:

```ts
    FormlyModule.forRoot({
      types: [
        {
          name: 'range',
          extends: 'input',
          defaultOptions: {
            templateOptions: {
              type: 'range',
              attributes: {
                min: 5,
                max: 10,
              },
            },
          },
        },
      ],
    }),
```

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ngx-formly/736)

### DIFF
--- a/src/core/src/components/formly.attributes.spec.ts
+++ b/src/core/src/components/formly.attributes.spec.ts
@@ -35,6 +35,9 @@ describe('FormlyAttributes Component', () => {
         expect(elm.getAttribute('name')).toBe('title-name');
         expect(elm.getAttribute('id')).toBe('title-id');
         expect(elm.getAttribute('aria-describedby')).toBe('title-id-message');
+        // using attributes option
+        expect(elm.getAttribute('min')).toEqual('5');
+        expect(elm.getAttribute('max')).toEqual('10');
         expect(fixture.componentInstance.field.focus).toBeFalsy();
       });
 
@@ -98,6 +101,10 @@ class TestComponent {
       'aria-describedby': true,
       tabindex: 5,
       step: 2,
+      attributes: {
+        min: 5,
+        max: 10,
+      },
     },
   };
 }
--- a/src/core/src/components/formly.attributes.ts
+++ b/src/core/src/components/formly.attributes.ts
@@ -33,6 +33,13 @@ export class FormlyAttributes implements OnChanges {
           this.elementRef.nativeElement, attr, this.getPropValue(this.field, attr),
         ));
 
+      if (this.field.templateOptions.attributes) {
+        const attributes = this.field.templateOptions.attributes;
+        Object.keys(attributes).forEach(name => this.renderer.setAttribute(
+          this.elementRef.nativeElement, name, attributes[name] as string,
+        ));
+      }
+
       this.statements
         .filter(statement => this.canApplyRender(fieldChanges, statement))
         .forEach(statement => this.renderer.listen(
--- a/src/core/src/components/formly.field.config.ts
+++ b/src/core/src/components/formly.field.config.ts
@@ -168,6 +168,7 @@ export interface FormlyTemplateOptions {
   pattern?: string|RegExp;
   required?: boolean;
   tabindex?: number;
+  attributes?: { [key: string]: string|number };
   step?: number;
   focus?: (field: FormlyFieldConfig, formControl: AbstractControl) => void;
   blur?: (field: FormlyFieldConfig, formControl: AbstractControl) => void;


juristr:
@aitboudad How would that work together with validations? Like if you want to specify both, a min & max + corresponding validator?

aitboudad:
`validator` is still applied here and there is no way to disable them right now.

juristr:
> `validator` is still applied here and there is no way to disable them right now.

Ya, I guessed so. Theoretically it doesn't hurt. But in general it would probably be much cleaner to have validators only inside the `validators` property. That would be a breaking change though.

You're adding the `attributes` array there to specify which of the `templateOptions` are actually valid attributes, right? Since we could specify other stuff there as well, which are not attributes. I see why you add an array, as you can maintain backwards compatibility, right?   
Because I'd nearly prefer a separate `attributes` object like...

```javascript
{
   templateOptions: {
      ...
      attributes: {
        min: 10,
        ...
     }
}
```

From an API perspective it would be clearer this way, what do you think? Or what does @beeman think 😉

aitboudad:
Changed to your proposal which seems mush cleaner, 
>Ya, I guessed so. Theoretically it doesn't hurt. But in general it would probably be much cleaner to have validators only inside the validators property. That would be a breaking change though.

maybe for a next major version `v3`.

juristr:
@aitboudad nice. Ya IMHO it's a breaking change as people might have used this already.

Given you're willing to do a breaking change. What do you think about the validators thing then? Like if we now add `min` to the `templateOptions` it would add a validator, right? As I mentioned previously, it would be more "intuitive" if validators are only added when specified in the `validators` property (given I understood properly how ngx-formly currently interprets them).

But I'm also wondering what @beeman says, as he's using formly also quite heavily

aitboudad:
> Given you're willing to do a breaking change.

I'm not sure yet but I would say that we allow several ways to set validation which may be confusing in some-cases and it should be improved.

  * [built-in validations](https://formly-js.github.io/ngx-formly/examples/validation/built-in-validations) (required, min, max, minLength, maxLength and pattern)
  * [validators](https://formly-js.github.io/ngx-formly/examples/validation/custom-validation) is most confused:
      - `validators.validation` which allow passing reactive forms validations or using alias `name`
      - `validators.[any key]` which allow passing `() => boolean` validator

beeman:
I am all for cleaning up the API in order to lessen the confusion for the users of Formly.

For me it makes total sense to move all validation properties to it's own key. 

I have written abstractions over Formly in several projects and most of the times the goal was to hide the implementation details, in order to create a more declarative API for defining forms.
 
Things I've abstracted:
- hide `templateOptions` (I personally think `options` is a better name).
- move validations to it's own an array 
- add helpers methods to quickly create inputs


```typescript
const emailDefaults = {
  label: 'Email',
  placeholder: 'Enter your email address',
}

export const field = (
  key: string, type: string = 'input', template: string = 'text', options: any = {}, validators: any = {}) => ({
  key,
  type,
  defaultValue: options.value || null,
  templateOptions: Object.assign({ type: template }, options),
  validators,
})

export const fieldEmail = (key: string, options: any = {}, validation: any = {}) => 
  field(key, 'input', 'email', Object.assign(emailDefaults, options), validation)
```

The abstraction above hides Formly's implementation details and allows to quickly create form fields without repeating yourself too much

```typescript
this.fields = [
  fieldEmail('myEmail'),
]
``` 


I realise this answers a broader question about Formly's API and we probably want to discuss this elsewhere. 

As far as this specific question about the min/max properties goes:
- I love the new `attributes` key as it saves me creating custom types
- I.m.o it makes a lot of sense to move field validators (min/max/required/maxLength) to the top-level
- We might be able to do it in a way that we log a deprecation warning to the console.

juristr:
> I'm not sure yet but I would say that we allow several ways to set validation which may be confusing in some-cases and it should be improved.

definitely, I agree that this should def be avoided.

> We might be able to do it in a way that we log a deprecation warning to the console.

I love that plan. that way we could maybe reroute the calls internally, allowing for a deprecation phase and making the breaking change less breaking 👍

aitboudad:
ok no breaking change, promised :), our codebase is pretty stable and we can focus on those improvements and documentation, each part should be discussed separately to figure out the best approach.

beeman:
Thanks for another quick merge @aitboudad.

I will open on an issue where we can discuss further API changes for a next major release.

PR #842, opened 2018-03-26
test(FormlyFormExpression): add test for service

samtsai:
Add unit tests for newly abstracted service.
See https://github.com/formly-js/ngx-formly/issues/711

Tests

**Other information**:

I'm fairly new to this library so I didn't create any stubs instead I copied over several setup functions from Form Builder. I tried to keep it at a minimum but I don't think it's very sustainable, so this PR is open for discussion :) Also, I haven't tested more complicated `FormGroup` scenarios like nested or `FormArray`.

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ngx-formly/842)

### DIFF
--- a/src/core/src/services/formly.form.expression.spec.ts
+++ b/src/core/src/services/formly.form.expression.spec.ts
@@ -8,7 +8,7 @@ import { evalStringExpression, evalExpressionValueSetter, FORMLY_VALIDATORS, get
 
 import { FormlyValueChangeEvent } from '../components/formly.field.config';
 import { FormlyConfig } from './formly.config';
-import { FormlyFieldConfig, FormlyFormBuilder, FormlyFormOptions, FormlyForm } from '../core';
+import { FormlyFieldConfig, FormlyFormBuilder, FormlyFormOptions, FormlyForm, FieldArrayType } from '../core';
 
 import { FormlyFormExpression } from './formly.form.expression';
 
@@ -20,6 +20,7 @@ describe('FormlyFormExpression service', () => {
 
   beforeEach(() => {
     TestComponent = MockComponent({ selector: 'formly-test-cmp' });
+
     expression = new FormlyFormExpression();
     form = new FormGroup({});
     builder = new FormlyFormBuilder(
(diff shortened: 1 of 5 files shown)


src/core/src/services/formly.form.expression.spec.ts
```diff
@@ -1,15 +1,281 @@
 import { FormlyConfig } from './formly.config';
-import { Validators, FormGroup } from '@angular/forms';
+import { FormlyFieldConfig, FormlyFormOptions } from '../core';
+import { AbstractControl, FormArray, FormControl, FormGroup, Validators } from '@angular/forms';
 import { Component } from '@angular/core';
+import { evalStringExpression, evalExpressionValueSetter, FORMLY_VALIDATORS, getKeyPath, isFunction, isObject } from './../utils';
 
 import { FormlyFormExpression } from './formly.form.expression';
 
+function registerFormControls(form: FormGroup | FormArray, fields: FormlyFieldConfig[], model: any, options: FormlyFormOptions) {
```
aitboudad: I'd prefer to use `FormlyFormBuilder` to build form instead

samtsai: One concern I had with using `FormlyFormBuilder` is that the public `buildForm` inherently calls FormlyFormExpression [`this.formlyFormExpression.checkFields(form, fields, model, options);`](https://github.com/formly-js/ngx-formly/blob/master/src/core/src/services/formly.form.builder.ts#L34).

So tests would look like this:

```
    builder.buildForm(form, fields, model, options);

    // expression.checkFields(form, fields, model, options); <-- inherently called but could be called explicitly called again

    expect(fields[1].hide).toBeTruthy();
    expect(fields[1].templateOptions.hidden).toBeTruthy();
```

As I'm typing this, what I could do, is initialize the FormBuilder and ignore it's internal instance of `formlyFormExpression` and have a separate instance just for the test.

Does this make sense? 😜

samtsai: But I agree I don't like having duplicate functions in the test itself.

aitboudad: you can pass a mocked instance instead to avoid call checkFields inside FormlyFormBuilder.

samtsai: Yes, I'll have something this afternoon! 👍

aitboudad:
Thank you @samtsai!

PR #853, opened 2018-03-30
Angular deps upgrade

### LINKED ISSUE #769: Upgrade Angular dependencies
I'm **currently working on this**. A PR will follow shortly

```
[x] feature request
[ ] support request
```

Currently there's otherwise an issue that on some machines (apparently) the examples page of the demo app doesn't start.

**Tasks**

- [x] Upgrade dependencies (currently working on [here](https://github.com/juristr/ng-formly/tree/angular-deps-upgrade))
- [ ] Adjust library packaging with `ng-packagr` (currently working on [here](https://github.com/juristr/ng-formly/tree/nx-workspace-setup))

juristr:

Updates Angular to the latest version. Moreover it adjusts ng-packagr.

@aitboudad Give it a look. Seems to work locally on my machine.

Thx for the hint about the patch. Took so long because I was working on a branch where to port this over to an NX workspace as there would already be everything setup to have libs and apps (for demo purposes here). However there's still [this issue on ng-packagr](https://github.com/dherges/ng-packagr/issues/86#issuecomment-375955829) which you intelligently circumvented with the patch 😉 .

Fixes #769

---
This change is [Reviewable](https://reviewable.io/reviews/formly-js/ngx-formly/853)


juristr:
What I was also wondering is whether the peer dependency on forms is still `>= 4.0.0`.  Theoretically it should, right? I'm not aware of any breaking changes there from v4 to v5.

aitboudad:
this will be part of the next major version so it should be `>= 5.0.0`, just changed the target branch into `next`

juristr:
But in theory there's no breaking change with `v4`, so by keeping the peer Dependency low we'd allow people to use the lib even if they're stuck on`v4`. I'd try to keep the peer dependency to Angular as low as possible.

Moreover, when do you expect `next` to land? Because this change actually is a prerequisite to run the demo app locally which otherwise doesn't work. As such it makes it hard to contribute :)

aitboudad:
libraries compiled with Angular 5 are not compatible with Angular 4.

> Because this change actually is a prerequisite to run the demo app locally which otherwise doesn't work. As such it makes it hard to contribute :)

next week probably I'll release `v2`, so I think all features should target `next` branch now.

aitboudad:
Thank you @juristr.

PR #916, opened 2018-04-26
docs(demo): add hide fields with animations example

aitboudad:


aitboudad:
need to update it into the latest version https://github.com/formly-js/ngx-formly/issues/800#issuecomment-386017066

aitboudad:
the funniest thing is I don't like angular animation at all :laughing:

juristr:
haha 😀. To be honest I didn't use them too much either, but those few times I had to use them they were quite easy to implement 🙂

PR #931, opened 2018-05-02
docs(guide) Update custom-formly-wrapper.md

thorgod:
Update to documents, added lifecycle to custom wrapper.

### DIFF
--- a/demo/src/app/guides/custom-formly-wrapper.md
+++ b/demo/src/app/guides/custom-formly-wrapper.md
@@ -154,3 +154,19 @@ Sometimes you always want a components with certain wrappers.
   export class AppModule { }
   ```
 You can do this by setting the `wrappers: ['fieldset', 'label']` to that type in the module
+
+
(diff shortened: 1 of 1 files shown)


demo/src/app/guides/custom-formly-wrapper.md
```diff
@@ -154,3 +154,19 @@ Sometimes you always want a components with certain wrappers.
   export class AppModule { }
   ```
 You can do this by setting the `wrappers: ['fieldset', 'label']` to that type in the module
+
+
+## Advanced Topics
+
+
+**life cycle**
+* [build field](https://github.com/formly-js/ngx-formly/blob/master/src/core/src/components/formly.form.ts#L61)
```
aitboudad: this info is not necessary a small summary is enough:
```markdown
**life cycle**

  1. pre-wrappers
  2. wrappers
  3. post-wrappers
  4. field component
```

PR #939, opened 2018-05-04
add Angular 6 support.

aitboudad:


angular.json
```diff
@@ -0,0 +1,375 @@
+{
```
thorgod: @aitboudad 
Did you manually have to convert your .angular-cli.json to this, or did the upgrade do it auto.

aitboudad: auto, `ng update @angular/cli` based on https://update.angular.io

samtsai:
Was going to ask about your thoughts on NG6 support but I knew if I just waited... 👍

PR #968, opened 2018-05-29
build: remove ng-packagr patch

aitboudad:


juristr:
Awesome, great it's now possible to define paths 👍.

PR #983, opened 2018-06-07
chore(schematics): fix schematics pkg version

samtsai:

Version fix for Schematics package.

Schematics published version is `0.0.1`.
Schematics has a different build process from the other `ngx-formly` packages and therefore needs a slightly modified build process.

To not alter too much, add a step to overwrite the dist `package.json` to have the "FORMLY_VERSION" key that is used to update to correct Formly version on builds.

**Other information**:

### DIFF
--- a/.config/version-schematics.js
+++ b/.config/version-schematics.js
@@ -1,0 +1,6 @@
+const fs = require('fs');
+// Schematics is NOT an `ng-package` and therefore requires a valid semver version
+const packagePath = 'dist/@ngx-formly/schematics';
+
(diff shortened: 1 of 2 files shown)


aitboudad:
The `0.0.1` was intentional :), the schematics release should be separated from the other packages for a moment (1 or 2 months) then it'll be sync with other packages.

samtsai:
@aitboudad makes sense!

aitboudad:
I'll revisit this PR when the time comes.

PR #1044, opened 2018-07-17
feat(core): remove lifecycle hooks in FieldType

aitboudad:
BREAKING CHANGE: lifecycle hooks in FieldType has been removed

in case of creating a custom field that extend `FieldType`, calling `super.{ngOnInit, ngAfterViewInit, ngOnDestroy}` has been removed

Before:
```ts
export class CustomFormlyField extends FieldType {
  ngOnInit() {
    super.ngOnInit();
    ...
  }
}
```

After:
```ts
export class CustomFormlyField extends FieldType {
  ngOnInit() {
    ...
  }
}
```


thorgod:
Just wondering why remove them? Where the events not being used anywhere?

PR #1078, opened 2018-07-30
fix(core): apply init fields type for fieldArray

aitboudad:
**What kind of change does this PR introduce? Bug fix**


aitboudad:
@enricooliva this may break your datatable example, the fix can be found in https://github.com/formly-js/ngx-formly/pull/1079

PR #1079, opened 2018-07-30
docs(demo): make columns configurable for ngx-datatable

### LINKED ISSUE #1076: ngx-datatable stackblitz old lib version / enhancements / bugs
```
[x] bug report
[x] enhancement
[x] support request
```

Thanks for your ngx-datatable implementation.

Opening ngx-datatable on Stackblitz shows on every row headers.
I think ngx-datatable support is only valid with ngx-formy/core 5.0.0 library.
On Stackblitz 4.6.6 is loaded. Changing it the 5.0.0-alpha library solved it.

I am mainly reporting this as I had a similar issue some months ago, which costs me a lot of time to identify a misbehaviour for the repeating section example and older libs.

Enhancement:
I adapted ngx-datatable example with an add and remove button, similar to repeating section example. Might be worth adding it to the example.
[https://stackblitz.com/edit/angular-wwqs8k?file=src%2Fapp%2Fdatatable.type.ts](https://stackblitz.com/edit/angular-wwqs8k?file=src%2Fapp%2Fdatatable.type.ts)
(FontAwesome buttons not showing correct and example needs some better layout)

Enhancement:
Repeating section and ngx-datatable formly type are derived from FieldArrayType. Similar to add/remove row method, is it possible to add a reset forms (model) method which can be called within the formly type component? 
Calling FormlyForms.resetModel() would reset the whole model and is not what I am looking for.

Bug:
Header should show an asterix on required fields.

Bug:
When adding a row in my above example new headers are created. Why? Bug?

aitboudad:
fix #1076

### DIFF
--- a/demo/src/app/examples/advanced/datatable-integration/app.component.ts
+++ b/demo/src/app/examples/advanced/datatable-integration/app.component.ts
@@ -41,13 +41,19 @@ export class AppComponent {
     {
       key: 'investments',
       type: 'datatable',
+      templateOptions: {
+        columns: [
+          { name: 'Name of Investment', prop: 'investmentName' },
+          { name: 'Date of Investment', prop: 'investmentDate' },
+          { name: 'Stock Identifier', prop: 'stockIdentifier' },
+        ],
+      },
       fieldArray: {
         fieldGroup: [
           {
             type: 'input',
             key: 'investmentName',
             templateOptions: {
-              label: 'Name of Investment:',
               required: true,
             },
           },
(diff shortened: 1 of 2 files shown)


enricooliva:
@aitboudad thank you.

For the label, I think the right solution is to use a particular wrapper to each field.

During the column generation I set "fieldset" wrapper. Is the type of wrapper that determine the label visualisation. May be we can define a new type of wrapper for the datable.

```
 this.field.fieldArray.fieldGroup.forEach(element => {
      element.wrappers = ['fieldset']
    });
```

enricooliva:
@aitboudad I tried to use sortable column. 

This is my working code:
```
 onSort(event) {
    const sort = event.sorts[0];
    this.formControl.value.sort((a , b) => {                
        if (typeof a[sort.prop] ===  "number"){
            return (a[sort.prop]>(b[sort.prop]) * (sort.dir === 'desc' ? -1 : 1));  
        }    
        return (a[sort.prop].localeCompare(b[sort.prop]) * (sort.dir === 'desc' ? -1 : 1));    
    });    
    this.formControl.patchValue(this.formControl.value);   
    this.model = [...this.formControl.value]; 
  }
```

PR #1093, opened 2018-08-07
feat(core): deprecate passing 'model' input to FormlyField

aitboudad:


thorgod:
What the benefit of removing the model from the FormlyField. 

I was using it for custom fields.

aitboudad:
less code :) and having one way to assign the model value to field, since the model value can be retrieved from the `field.model` I don't see any relevant use-case to keep it.

PR #1111, opened 2018-08-09
feat(core): rename Field to FieldType

aitboudad:


thorgod:
Great job on all the new changes! I understand the name change, its hard keeping up with all the changes making sure my code that uses core features does not break. I know this change does break anything just something else I will need to fix.

Overall I like the changes, just be aware that we rely on some of the core functionality

aitboudad:
> Overall I like the changes, just be aware that we rely on some of the core functionality

then give it a try with `v5.0.0-beta.2` before it's too late :).

PR #1125, opened 2018-08-14
Ionic

Devqon:
Feature: Extend Ionic v4 support

Limited support for ionic options

- general
  - moved `ion-item` wrapper to wrapper element
  - support for `ionChange` event on all types
- ion-select
  - support for interface option

### DIFF
--- a/src/ionic/src/lib/ui-ionic.config.ts
+++ b/src/ionic/src/lib/ui-ionic.config.ts
@@ -47,7 +47,7 @@ export const IONIC_FORMLY_CONFIG: ConfigOption = {
     {
       name: 'radio',
       component: FormlyFieldRadio,
-      wrappers: ['form-field'],
+      wrappers: [],
     },
     {
       name: 'toggle',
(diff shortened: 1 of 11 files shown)


src/ionic/src/lib/ui-ionic.config.ts
```diff
@@ -57,7 +57,7 @@ export const IONIC_FORMLY_CONFIG: ConfigOption = {
     {
       name: 'select',
       component: FormlyFieldSelect,
-      wrappers: ['form-field'],
+      wrappers: [], // TODO: find out why the wrapper doesn't work with an ion-select
```
aitboudad: for select add the following style in `FormlyFieldSelect`:
```ts
@Component({
  selector: 'formly-field-ion-select',
  ...,
  styles: [':host { display: inherit; }'],
})
export class FormlyFieldSelect extends IonFieldType {
}
```

Devqon: Awesome thanks!

Devqon:
Question: should I add *all* ionic functionality that is available into the templates? For example, [the select](https://beta.ionicframework.com/docs/api/select#properties) can take more arguments, should I add all of them to the template?

aitboudad:
why not if it's necessary but I think it should be addressed in another PR

aitboudad:
Thank you @Devqon!

PR #1137, opened 2018-08-21
enabled bootstrap checkbox/multicheckbox to hide label when specifically requested

ntziolis:
It's pretty straight forward

### DIFF
--- a/src/bootstrap/src/lib/types/checkbox.ts
+++ b/src/bootstrap/src/lib/types/checkbox.ts
@@ -10,7 +10,7 @@ import { FieldType } from '@ngx-formly/core';
         [indeterminate]="to.indeterminate && (model || {})[key] === undefined"
         [formControl]="formControl"
         [formlyAttributes]="field">
-      <label class="custom-control-label" [for]="id">
+      <label *ngIf="to.label && !to.hideLabel" class="custom-control-label" [for]="id">
         {{ to.label }}
         <span *ngIf="to.required && to.hideRequiredMarker !== true">*</span>
       </label>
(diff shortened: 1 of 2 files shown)


src/bootstrap/src/lib/types/checkbox.ts
```diff
@@ -10,7 +10,7 @@ import { FieldType } from '@ngx-formly/core';
         [indeterminate]="to.indeterminate && (model || {})[key] === undefined"
         [formControl]="formControl"
         [formlyAttributes]="field">
-      <label class="custom-control-label" [for]="id">
+      <label *ngIf="to.label && !to.hideLabel" class="custom-control-label" [for]="id">
```
aitboudad: use different name since `hideLabel` is used in the main wrapper:
``` 
*ngIf="to.label && to.hideCheckboxLabel !== true"
```

ntziolis: Done

src/bootstrap/src/lib/types/multicheckbox.ts
```diff
@@ -12,7 +12,7 @@ import { Observable } from 'rxjs';
         [value]="option.value"
         [formControl]="formControl.get(option.key)"
         [formlyAttributes]="field">
-      <label class="custom-control-label" [for]="id + '_' + i">
+      <label *ngIf="to.label && !to.hideCheckboxLabel" class="custom-control-label" [for]="id + '_' + i">
```
mohammedzamakhan: shouldn't this be `*ngIf="option.value && !to.hideCheckboxLabel"`

aitboudad:
the last time I checked this I noticed that the checkbox is entirely hidden, to hide label we should follow https://getbootstrap.com/docs/4.0/components/forms/#without-labels which I'm not in favor it right now.

sorry :pray: !

PR #1155, opened 2018-08-29
feat(core): deprecate fieldTransform in favor of custom extension

aitboudad:


rdev-software:
fieldTransform is this in documentation, but not documented ? How to use it, or what is alternative, if its deprecated ?

aitboudad:
> what is alternative

a custom extension which allows dealing with field through the following methods:
* `prePopulate`: before touching the field.
* `onPopulate`:  defaultOptions, Model, formControl are set (which is suitable to update the child element).
* `postPopulate`: after populate children, if the field doesn't have children it should be the same as `onPopulate` event.

**Note**: These events can be called whenever the field model change not matter the field was initialized or not.

> How to register an extension?

```ts
export function prePopulateFieldHooks(field: FormlyFieldConfig) { ... }
export const FieldHooksExtension: FormlyExtension = {
  prePopulate: prePopulateFieldHooks
}

@NgModule({
  imports: [
    ...,
    FormlyModule.forRoot({
      ...,
      extensions: [{ name: 'custom_name', extension: FieldHooksExtension }],
    }),
  ],
})
export class AppModule { }
```

check the following examples which rely on extension:

- https://ngx-formly.github.io/ngx-formly/examples/advanced/i18n-alternative
- https://ngx-formly.github.io/ngx-formly/examples/other/material-prefix-suffix

rdev-software:
Sure but these extensions does not have method "onUpdate", so yes, you can make extension to "parse" stuff, but for that you have parsers already. We are missing formatters, ie way to show things differently in custom component (or existing components) without actually touching model. 

For now I've created custom controller, which on every ngDoCheck() transform "model" to desired value = works like "formatter" and then on update, translating result back to actual model, so works like "parser".

But thats just too "hacky" and its not a way how dynamic forms should work. I used https://github.com/xmartlabs/Eureka in iOS and that gives you full control over "view" of your own "model"

aitboudad:
could you please show me in a new issue an example of what you've done to have an idea maybe I'll have a better way to deal with it, thanks!

rdev-software:
Every single attempt I am trying I will end up using custom control, and I dont think thats the purpose of dynamic forms. Thats all. Lets see this simple example

I have value "difficulty", it can have value from 0-1 (double)

So I setup simple range control:
```
  {
      key: 'difficultyFactor',
      type: 'range',
      templateOptions: {
        label: 'Mission Difficulty',
        min: 0,
        max: 1,
      },
    },
```

Of course this does not work, as ionic component takes only integer numbers...so I need
```
  {
      key: 'difficultyFactor',
      type: 'range',
      templateOptions: {
        label: 'Mission Difficulty',
        min: 0,
        max: 100,
      },
    },
```

But manipulate with model using formatter & parser, parser works perfectly fine with
```
 parsers: [this.toPecent],

toPecent(value) {
    return value / 100.0;
  }
```

But now when I want to read value, slider is taking my value of 0.5 and always draw it on top left...as my model never exceed "1".

So I need formatter to actually multiply result by 100, before each update.

Please dont tell me that I am first one trying to do that in version 5.1.2 ? @aitboudad

aitboudad:
> Every single attempt I am trying I will end up using custom control

then share your concern and we'll provide a solution.

> I dont think thats the purpose of dynamic forms

take it easy, we're trying our best based on our free time to provide a such work (we don't earn anything from this!) so I'll ignore that one.

> Of course this does not work, as ionic component takes only integer numbers...so I need

use `step`:

```patch
    {
      key: 'difficultyFactor',
      type: 'range',
      templateOptions: {
        label: 'Mission Difficulty',
        min: 0,
        max: 1,
+       step: 0.1,
      },
    },
```

> Please dont tell me that I am first one trying to do that in version 5.1.2 ?

you may ask https://github.com/angular/angular/issues/3009, if you want a solution form me fill an issue gently and I'll try to find one when time allows!

rdev-software:
@aitboudad thx for that, sorry about my comments...I've pushed for dynamic forms in solution We are building and I am under too much pressure. Its not your fault at all. 

Its just I am sure that there are solutions for simple things I am trying to achieve and I am frustrated that I cant find answers myself and have to write comments and bother your personally :(

PR #1182, opened 2018-09-18
feat(bootstrap): use ngValue for option with 'null' for placeholder

### LINKED ISSUE #1169: Make the bootstrap select placeholder visible
```
[x] feature request
[x] support request
```

**Current behavior**
When specifying the placeholder for the select UI component it is only visible on click event.  

**Expected behavior**
The select placeholder should be visible without the need to click (like the placeholder of the input field type). 

An example of the behaviour I'm asking for is show at this [stackblitz](https://stackblitz.com/edit/angular-koejw2).

Even if it is simple to define a customized select template (as shown in the linked stackblitz code), I think it would be easier to update the "select" types of the library. 
For example, for the bootstrap select type you just need to substituting 

> `<option *ngIf="to.placeholder" value="">{{ to.placeholder }}</option>`

with 

> `<option *ngIf="to.placeholder" [ngValue]="null">{{ to.placeholder }}</option>`

aitboudad:
fix #1169


antonmakarenko:
@aitboudad Is it possible to backport this fix to 4.8.x branch?

aitboudad:
nop, either upgrade to [V5](https://github.com/ngx-formly/ngx-formly/blob/master/UPGRADE-5.0.md) or use a custom select type which override the default one.

PR #1190, opened 2018-09-20
fix(bootstrap): get correct radio value through queryselector

### LINKED ISSUE #1187: The "value" property on radio buttons isn't updated
```
[ x ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request
```

**Current behavior**
I'm using bootstrap

When I query a radio input element, for example: `myRadioInput` and check it's value like `myRadioInput.value` I get "on" instead of the actual value from "option.value"

**Expected behavior**
I expect that `myRadioInput.value` will be the `option.value` I set on the options array.

In the following link:
1. Open the console 
2. Click on the button
3. See that instead of "male" it says "on"

[https://stackblitz.com/edit/ngx-formly-ui-bootstrap-radio-value-bug](https://stackblitz.com/edit/ngx-formly-ui-bootstrap-radio-value-bug)

For testing purposes, I need to be able to click on a certain radio input to manually select it.
that's why it's important to be able to compare the radio.value to the manually selected value. 

windows 10
yarn
Formly + formly-bootstrap versions: 4.7.2

* **Angular version:** 6.1.0

* **Browser:** [ Chrome 68.0.3440.106  ]

* **Language:** [TypeScript 2.9 ]

* **Node (for AoT issues):** `node --version` =

aitboudad:
fix #1187

### DIFF
--- a/src/bootstrap/src/lib/types/radio.ts
+++ b/src/bootstrap/src/lib/types/radio.ts
@@ -10,6 +10,7 @@ import { FieldType } from '@ngx-formly/core';
           [id]="id + '_' + i"
           [name]="id"
           [class.is-invalid]="showError"
+          [attr.value]="option.value"
           [value]="option.value"
           [formControl]="formControl"
           [formlyAttributes]="field">


Nielsb85:
I know this is old, but the material version seems to have the same problem

aitboudad:
@Nielsb85 could you please fill a new issue with a reproduction example. Thanks!

PR #1207, opened 2018-09-27
feat: add templateOptions.size propery support

alQlagin:
Added support input element sizing 
Described in https://getbootstrap.com/docs/4.1/components/forms/#sizing


aitboudad:
you may check my comment https://github.com/formly-js/ngx-formly/issues/1193#issuecomment-423933302 for a general solution

PR #1299, opened 2018-11-17
The same ValidationErrors wants to prompt different information, based on the value entered

smile2014:
Chinese ID card 15 and 18 valid check
export function  cardValidator(control: FormControl): ValidationErrors {
  let message ="";
   if(card === '')
    {
        message = 'card can't is not null';
        return false;
    }
    if(isCardNo(card) === false)
    {
        message='card can't correct';
        return false;
    }
    if(checkProvince(card) === false)
    {
        message='card can't correct';
        return false;
    }
    if(checkBirthday(card) === false)
    {
        message='card can't correct';
        return false;
    }
}

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+<a name="4.8.1"></a>
+## [4.8.1](https://github.com/formly-js/ngx-formly/compare/v4.8.0...v4.8.1) (2018-11-03)
+
+
+### Bug Fixes
+
+* **material:** set select aria-labelledby to avoid ExpressionCheck Error ([#1265](https://github.com/formly-js/ngx-formly/issues/1265)) ([e12b6b9](https://github.com/formly-js/ngx-formly/commit/e12b6b9)), closes [#1251](https://github.com/formly-js/ngx-formly/issues/1251)
+
+
+
 <a name="4.8.0"></a>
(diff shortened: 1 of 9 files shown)


smile2014:
@aitboudad  Can write multiple valid, but that's too much trouble 
Can you give me some advice, thank you?

aitboudad:
sorry, but your question is not clear to me!

aitboudad:
the message can be passed as a function see https://formly-js.github.io/ngx-formly/examples/validation/validation-message

smile2014:
I know, but the same processor can't prompt different messages.
For example
https://github.com/ScottMGerstl/ng2-validate/blob/master/demo/email.validator.ts

smile2014:
@aitboudad 
This configuration allows erros to read this message and find that ngx-formly does not support this.

export function IpValidator(control: FormControl): ValidationErrors {
  return !control.value || /(\d{1,3}\.){3}\d{1,3}/.test(control.value) ? null : { 'ip': true,message:'testerro' };
}

aitboudad:
we don't support that case yet, just opened an issue for that #1303

smile2014:
@aitboudad  thank you ，the problem has been solved
export function IpValidator(control: FormControl): ValidationErrors {
return !control.value || /(\d{1,3}.){3}\d{1,3}/.test(control.value) ? null : { 'ip': {message:'customMessage'} };
}

aitboudad:
seems already supported [FormlyValidationMessage:L27](https://github.com/formly-js/ngx-formly/blob/master/src/core/src/lib/templates/formly.validation-message.ts#L27) :), I'll try to add it in the demo example.

smile2014:
@aitboudad thank you

PR #1325, opened 2018-11-30
fix(material): use bindings instead of formlyAttributes in material

### LINKED ISSUE #1317: Material with formlyAttributes
```
[ ] feature request
[x] support request
```

Every now and then I observe a strange minor glitch in how material components with ngx-formly. Some of them were already reported(#1281, #1277, #1276).

---

Latest issue I found is that `tabindex` is probably not provided to material components as expected.
Material checks for `tabindex` attribute, it does not have `tabindex` Input:

```
  constructor(elementRef: ElementRef<HTMLElement>,
              private _changeDetectorRef: ChangeDetectorRef,
              private _focusMonitor: FocusMonitor,
              private _ngZone: NgZone,
              @Attribute('tabindex') tabIndex: string, <<<<<-------- HERE
              @Optional() @Inject(MAT_CHECKBOX_CLICK_ACTION)
                  private _clickAction: MatCheckboxClickAction,
              @Optional() @Inject(ANIMATION_MODULE_TYPE) public _animationMode?: string) {
```

This means, that tabindex provided by `FormlyAttributes` does not reach this constructor:

```
host: {
    '[attr.name]': 'field.name',
    '[attr.placeholder]': 'to.placeholder',
    '[attr.tabindex]': 'to.tabindex || 0', <<<<-------- HERE
    '[attr.readonly]': 'to.readonly',
    '[attr.step]': 'to.step',
```

I am not 100% sure, but we might expect some issues because of that, since we are setting tab index in a way component does not expect.

Simplified demo of `[attr.tabindex]` and `@Attribute('tabindex')`: https://stackblitz.com/edit/angular-tbqg1t?file=src%2Fapp%2Fattr.directive.ts

---

In order to avoid all these issues it maybe makes sense not to use `[attr.]` for Material components, but add them directly to the component?

This will however not solve the `tabindex` issue, since Material seems to expect hardcoded value there?

klemenoslaj:
HostBinding attributes like readonly are not respected by Material:
https://github.com/angular/material2/issues/14024

fix #1317

Bug fix

https://github.com/formly-js/ngx-formly/issues/1317

Bottom line, `@angular/material` is not compatible with `formlyAttributes`, therefore we have to change it to use input bindings (_just for `@ngx-formly/material`_).  
As well material specific `FormlyAttributes` was created(_`FormlyMatAttributes`_), which does not add the following three attributes, because they should be always added as an inputs: `placeholder`, `tabindex`, `readonly`.

**Other information**:
Considering the changes made, unit test does not make much sense.  
If there is a case where unit test should be implemented please let me know.


src/material/form-field/src/formly.attributes.ts
```diff
@@ -0,0 +1,19 @@
...
+    '[attr.step]': 'to.step',
+
+    '(keyup)': 'to.keyup && to.keyup(field, $event)',
+    '(keydown)': 'to.keydown && to.keydown(field, $event)',
+    '(click)': 'to.click && to.click(field, $event)',
+    '(change)': 'to.change && to.change(field, $event)',
+    '(keypress)': 'to.keypress && to.keypress(field, $event)',
+  },
+})
+export class FormlyMatAttributes extends FormlyAttributes {
```
aitboudad: I would prefer to keep `formlyAttributes` (we can adjust it to make it  compatible with material if necessary), since I'm thinking that most the material issues would be gone on `ivy` implementation,

klemenoslaj: In fact I tried to set all attributes to `null` in material use case, but if any of mentioned attributes is present in `host:`, certain components will have some edge case issues.

Example:
- Add `'[attr.tabindex]': 'to.tabindex || 0',` back to `formlyMatAttributes`
- Run demo, navigate to Material and scroll to checkbox
- Set checkbox to `false` value
-- Now you should be getting `mat-error`, but you won't

---

Probably the only way to make this work is to add attributes with `Renderer2` and add an option to disable it for Material?

klemenoslaj: I am just trying it out `Renderer2` and it seems to be working fine, tests are passing as well.

---

Injection:
```
@Optional() @Attribute('material') private isMaterial: boolean,
```

In `ngOnChanges` livecycle hook:
```
if (!this.isMaterial) {
  this.renderer.setAttribute(this.elementRef.nativeElement, 'placeholder', this.to.placeholder);
    this.renderer.setAttribute(this.elementRef.nativeElement, 'tabindex', `${this.to.tabindex || 0}`);
  this.renderer.setAttribute(this.elementRef.nativeElement, 'readonly', this.to.readonly);
}
```

---

I am however not happy with `material` attribute..Do you have any other idea how to detect if directive is used within a material?

One option would be to again extend the class and set the property `isMaterial` to `true`, by default it's false...don't like it either.

aitboudad: can you try adding those attributes as inputs:
```ts
@Directive({
  selector: '[formlyAttributes]',
  host: {
    '[attr.tabindex]': 'tabindex ? tabindex : to.tabindex',
  },
})
export class FormlyAttributes implements OnChanges {
  @Input() tabindex;
}
```

klemenoslaj: @aitboudad brilliant, should have think of that first.
I would however prefer the following  syntax: `'[attr.placeholder]': 'placeholder || to.placeholder',`.

What do you think?

aitboudad: seems better :)

klemenoslaj: Will do a bit of testing over the known issues of material and then will push.

klemenoslaj: Just realised I forgot to replace `formlyMatAttributes` with `formlyAttributes`, so I was basically testing the current workaround. Sorry for that.

So the following code, sadly does not work:
```
@Directive({
  selector: '[formlyAttributes]',
  host: {
    '[attr.name]': 'field.name',
    '[attr.placeholder]': 'placeholder || to.placeholder',
    '[attr.tabindex]': 'tabindex || to.tabindex || 0',
    '[attr.readonly]': 'readonly || to.readonly',
    '[attr.step]': 'to.step',

    // ...
  },
})
export class FormlyAttributes implements OnChanges {
  @Input('formlyAttributes') field: FormlyFieldConfig;
  @Input('placeholder') placeholder: number;
  @Input('tabindex') tabindex: number;
  @Input('readonly') readonly: number;

  // ...
```

Seems like putting the attribute like that on some material components really brings in hard to notice side effects.

I will try to find a cleaner solution than `@Attribute` suggested above.

klemenoslaj: Ok I just figured that adding the attributes using `Renderer2` does not trigger any of the edge cases I was able to indicate earlier. So I would suggest to remove attributes from `host:` and instead always add them with `Renderer2`. This I don't consider an ugly solution, since we use `Renderer2` already anyway.

Like this:
```

if (changes.field) {
  const fieldChanges = changes.field;

  this.renderer.setAttribute(this.elementRef.nativeElement, 'id', this.field.id);
  this.renderer.setAttribute(this.elementRef.nativeElement, 'placeholder', this.to.placeholder);
  this.renderer.setAttribute(this.elementRef.nativeElement, 'tabindex', `${this.to.tabindex || 0}`);
  this.renderer.setAttribute(this.elementRef.nativeElement, 'readonly', `${this.to.readonly}`);
```

> _If you wish to reproduce it, serve the demo with `aot` and try to put a checkbox  example in material UI to false - error should appear_

Are you ok with this solution?

klemenoslaj: Yes, that's true. Basically the changes to `templateOptions` from the outside would not be  reflected on a controls, which people might find strange and file another bug report.

What about ngDoCheck  live cycle hook with a dirty checking?

aitboudad: I'd prefer to wait until checking on my side tomorrow, to see if there is a simple way

klemenoslaj: Deal :+1:
Will throw the working code with `ngDoCheck` here anyway:

```
ngDoCheck() {
  if (this.placeholder !== this.to.placeholder) {
    this.renderer.setAttribute(this.elementRef.nativeElement, 'placeholder', this.to.placeholder);
    this.placeholder = this.to.placeholder;
  }

  if (this.tabindex !== this.to.tabindex) {
    this.renderer.setAttribute(this.elementRef.nativeElement, 'tabindex', `${this.to.tabindex || 0}`);
    this.tabindex = this.to.tabindex;
  }

  if (this.readonly !== this.to.readonly) {
    this.renderer.setAttribute(this.elementRef.nativeElement, 'readonly', `${this.to.readonly}`);
    this.readonly = this.to.readonly;
  }
}
```

klemenoslaj: @aitboudad found any better way so far?

aitboudad: nop, we can go with that approach for now.

klemenoslaj: Ok, I will push my change then 👍

aitboudad:
Thank you @klemenoslaj!

PR #1333, opened 2018-12-06
fix(core): ensure init options during buildForm

aitboudad:


juristr:
That was fast 👍🙂. Commenting here, so once it's merged I'll test it on my setup

PR #1372, opened 2019-01-10
fix(material): remove duplicated readonly prop

juristr:

Quick fix, just removing a duplicate setting of the `readonly` prop on the material input type. doesn't create any side effect.


aitboudad:
> Guess it should be added right @aitboudad ?

I think so, do it in separate PR!

juristr:
> I think so, do it in separate PR!

Alright, will send you one on Monday 👍

PR #1409, opened 2019-02-03
chore(demo): avoid relying on FormlyForm::ng-content

aitboudad:


juristr:
@aitboudad I actually never placed something inside the formly form tag and always wondered what it was good for. did it have any special meaning/functionality?

aitboudad:
@juristr it was useful before https://github.com/ngx-formly/ngx-formly/pull/93 (which move `form` tag outside `formly-form`) but since then I don't see any relevant use-case.

juristr:
ah I see, makes sense. in fact assumed it had to do with such scenario 👍

Issue #1398, opened 2019-01-28
[Question] What is the `formly` way to remove a form field and its FormControl(s)?


samtsai:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[X] support request
```

**Current behavior**

I am dynamically adding fields based on User selections on previous form fields.  All of this works as expected, but we allow the User to remove these fields (e.g. similar to repeater examples).  What I cannot figure out is how to properly do a full clean up.

I currently manually handle it by removing each part individually: field from `FormlyFieldConfig[]`, formControl from the form, `null` or deleting the property off of the model.  This process seems a little too manual and I wanted to know if there's a more `formly` way to do this.  See https://stackblitz.com/edit/ngx-formly-ui-material-i9sndp

Note: I'm not looking to using `hide` or `hideExpression` for these fields, since I would rather completely remove them.

aitboudad:
we only deal with repeatable field, so to remove a form field you have to get the field and remove its formControl and model.
dealing with it on our side would produce a breaking change which I'm not sure it worth it right now.

aitboudad:
let's give it a try!

aitboudad:
I propose adding a new option `strict` which force immutability and cleanup the removed field (formControl and model value).

`strict` = immutable( #1427) + cleanup the formcontrol and its model value.

WDYT ?

samtsai:
I generally like immutability pattern, so I think adding `strict` is reasonable!

aitboudad:
added to `V6` roadmap

PR #1427, opened 2019-02-18
fix(core): cleanup fields on destroy

### LINKED ISSUE #1426: Converting circular structure to JSON
```
[ ] feature request
[ X] support request
```

Hi,
I'm using formly inside a "Matdialog", and I found something strange.
When I open the dialog for the second time the mandatory fields are already "touched",
and if i try to stringify the form fields (the second time) I get this error 
ERROR
Error: Converting circular structure to JSON

I can not understand where I'm wrong.

https://stackblitz.com/edit/angular-3rd8sk-ytzwep

aitboudad:
fix #1426


samtsai:
Linking to #1398 in case it helps with that

aitboudad:
@samtsai well not related to that issue at all :), built `form` still untouched,
your request is valid but it shouldn't be the default behavior (at least for now)

aitboudad:
I'm going to revert this in favor of #1430 which resolve the cleanup issue without a lot of extra core!

PR #1430, opened 2019-02-19
feat(core): add support for immutable mode

### LINKED ISSUE #1429: Weird behavior when using the same field config for different forms on the same page
```
[ x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request
```

**Current behavior**
Using the same keys for different separate forms on the same page formly could trigger an error under some circumstances:
```
Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'disabled: false'. Current value: 'disabled: true'
``` 
This happens in an application using angular material when I have one form open on the page and then open the other instance inside an modal.

If I create an copy (via `JSON.parse(JSON.stringify(val))`) of the config inside the modal I get a `onverting circular structure to JSON` error. If I create a copy at both of the instances everything runs fine.

**Expected behavior**
Using the same config for different forms should not lead to errors.

While so far unable to create a blitz which reproduces the exact issue I am having, I'll made another one: https://stackblitz.com/edit/ngx-formly-ui-material-fngtya

If you click on the checkbox of the second form, the checkbox of the first form is focused. I am assuming that this connection leads to the error.

It should be possible to reproduce the described behavior exactly in this repo: https://github.com/johannesjo/super-productivity/commit/1fab8ef2df92f8004dc364d61a560e8c9149ec50#commitcomment-32363809

* **Angular version:** 7.2.X

* **Browser:** all (I guess)

* **Language:** all (I assume)

I'll try to dig a little deeper. Thanks for this very cool module.

aitboudad:
fix #1429

### DIFF
--- a/src/core/src/lib/components/formly.form.ts
+++ b/src/core/src/lib/components/formly.form.ts
@@ -1,8 +1,8 @@
-import { Component, DoCheck, OnChanges, Input, SimpleChanges, Optional, EventEmitter, Output, SkipSelf, OnDestroy, ComponentFactoryResolver, Injector } from '@angular/core';
+import { Component, DoCheck, OnChanges, Input, SimpleChanges, Optional, EventEmitter, Output, SkipSelf, OnDestroy, ComponentFactoryResolver, Injector, Attribute } from '@angular/core';
 import { FormGroup, FormArray, NgForm, FormGroupDirective } from '@angular/forms';
 import { FormlyFieldConfig, FormlyFormOptions, FormlyFormOptionsCache } from './formly.field.config';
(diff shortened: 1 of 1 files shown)


samtsai:
Can you post a demo of this in use?  I've just added the attribute to one of my forms and I pretty much got a memory leak.  It's possible that I need the side effects of not immutability the way that I set up my forms.

I'll eventually try to do a stackblitz demo of the memory leak.

aitboudad:
just pushed a fix in `5.0.0-rc.6`, give it a try and let me know if it's working

> Can you post a demo of this in use?

https://stackblitz.com/edit/angular-vjavfb

Issue #1459, opened 2019-03-07
@ngx-formly/json-schema


juristr:
##  1. Progress of supported schema:

- [x] string
  - [x] Length (`minLength`, `maxLength`)
  - [x] Regular Expressions (`pattern`)
  - [ ] Format
- [x] Numeric types
  - [x] integer
  - [x] number
  - [x] Multiples (`multipleOf`)
  - [x] Range (`minimum`, `exclusiveMinimum`, `maximum` and `exclusiveMaximum`)
    - [x] `minimum` and `maximum`
    - [x] `exclusiveMinimum` and `exclusiveMaximum`
- [x] object
  - [x] Properties
  - [x] Required Properties
  - [ ] Property names
  - [ ] Size
  - [x] Dependencies
  - [ ] Pattern Properties
- [x] array
  - [x] Items
  - [x] Length
  - [x] Uniqueness
  - [ ] contains
- [x] boolean
- [x] null
- [x] Generic keywords
  - [x] Annotations (`title`, `description`, `default`)
  - [x] Enumerated values
  - [x] Constant values
- [ ] Media: string-encoding non-JSON data
  - [ ] contentMediaType
  - [ ] contentEncoding
- [ ] Combining schemas
  - [x] allOf
  - [x] anyOf
  - [x] oneOf
  - [ ] not
- [ ] Applying subschemas conditionally (`if`, `then` and `else`)
- [x] Structuring a complex schema
  - [x] $ref
  - [ ] $id
- [ ] extra options
    - [x] widget
    - [x] map
    - [ ] uiSchema
    - [ ] convert formly schema to json-schema

##  3.Docs
  - [ ] add guide section
  - [ ] improve demo examples
  - [ ] options
    - [ ] widget
    - [ ] map
    - [ ] uiSchema

juristr:
@aitboudad @kenisteward I've followed your issue #1348 and I'd like to start developing on some of those things (as mentioned in my original issue description), specifically I'd need

- dynamic select boxes/autocompletes (i.e. where the data is fetched on the fly by giving it a source URL to the API)
- datepickers

And for those, the main issues I'm currently encountering is

1. where to define the "widget" to use, i.e. `<select>` or some custom `autocomplete`, basically what would then translate to some formly type.
1. where to put custom attributes, like the `source` (or how you'd call it) property that specifies the URL where to fetch the entries for the select box

I started giving a look to:

- https://mozilla-services.github.io/react-jsonschema-form/
[...]

PR #1467, opened 2019-03-13
feat(core): deep merge widget prop & add map callback

juristr:

As discussed in #1459 this adds 2 things

**deep merge of `widget.formlyConfig` property using the already existing algorithm

```
...
// map in possible formlyConfig options from the widget property
if (jsonSchema['widget'] && jsonSchema['widget'].formlyConfig) {
  const widgetConfig = jsonSchema['widget'].formlyConfig;
  field = reverseDeepMerge(field, widgetConfig);
}
...
```

The second thing is **the ability to pass in a `FormlyJsonschemaOptions` with a `map` function callback that allows to further customize the mapping of the formly field from the JSONSchema source.

**Other information**:

@aitboudad There is an open question here, namely whether the "deep merge" with the `widget.formlyConfig` object should overwrite existing mappings or not. Right now I'm doing the mappings (you already had in the `FormlyJsonschema` and then I apply the "deep merge" by calling the `reverseDeepMerge`. I've seen that function doesn't overwrite, so I left this behavior for now.
Still thinking whether that's right 🤔.

Because that also means that something like

```
{
   type: 'integer',
   ...,
   widget: {
      formlyConfig: {
         type: 'select',
         templateOptions: { ... }
      }
   }
}
```

wouldn't work. The `type` is already set by the JSONSchema so the `formlyConfig` doesn't overwrite it. What do you think?

### DIFF
--- a/src/core/json-schema/src/formly-json-schema.service.ts
+++ b/src/core/json-schema/src/formly-json-schema.service.ts
@@ -1,15 +1,25 @@
 import { Injectable } from '@angular/core';
 import { FormlyFieldConfig } from '@ngx-formly/core';
 import { JSONSchema7, JSONSchema7TypeName } from 'json-schema';
+import { ɵreverseDeepMerge as reverseDeepMerge } from '@ngx-formly/core';
+
+export interface FormlyJsonschemaOptions {
+  /**
+   * allows to intercept the mapping, taking the already mapped
+   * formly field and the original JSONSchema source from which it
+   * was mapped.
+   */
+  map?: (mappedField: FormlyFieldConfig, mapSource: JSONSchema7) => FormlyFieldConfig;
+}
 
 @Injectable({ providedIn: 'root' })
 export class FormlyJsonschema {
-  toFieldConfig(jsonSchema: JSONSchema7): FormlyFieldConfig {
-    return this._toFieldConfig(jsonSchema);
+  toFieldConfig(jsonSchema: JSONSchema7, options?: FormlyJsonschemaOptions): FormlyFieldConfig {
+    return this._toFieldConfig(jsonSchema, null, options);
   }
 
-  _toFieldConfig(jsonSchema: JSONSchema7, key?: string): FormlyFieldConfig {
-    const field: FormlyFieldConfig = {
+  _toFieldConfig(jsonSchema: JSONSchema7, key?: string, options?: FormlyJsonschemaOptions): FormlyFieldConfig {
+    let field: FormlyFieldConfig = {
       ...(key ? { key } : {}),
       type: jsonSchema.type as JSONSchema7TypeName,
       defaultValue: jsonSchema.default,
@@ -39,7 +49,7 @@ export class FormlyJsonschema {
       case 'object': {
         field.fieldGroup = [];
         Object.keys(jsonSchema.properties).forEach(p => {
-          const child = this._toFieldConfig(<JSONSchema7> jsonSchema.properties[p], p);
+          const child = this._toFieldConfig(<JSONSchema7> jsonSchema.properties[p], p, options);
           if (Array.isArray(jsonSchema.required) && jsonSchema.required.indexOf(p) !== -1) {
             child.templateOptions.required = true;
           }
@@ -49,13 +59,13 @@ export class FormlyJsonschema {
       }
       case 'array': {
         if (!Array.isArray(jsonSchema.items)) {
-          field.fieldArray = this._toFieldConfig(jsonSchema.items as JSONSchema7);
+          field.fieldArray = this._toFieldConfig(jsonSchema.items as JSONSchema7, key, options);
         } else {
           field['_fieldArray'] = [];
           field.fieldGroup = [];
-          jsonSchema.items.forEach(item => field['_fieldArray'].push(this._toFieldConfig(<JSONSchema7> item)));
+          jsonSchema.items.forEach(item => field['_fieldArray'].push(this._toFieldConfig(<JSONSchema7> item, key, options)));
           if (jsonSchema.additionalItems) {
-            field['_additionalFieldArray'] = this._toFieldConfig(<JSONSchema7> jsonSchema.additionalItems);
+            field['_additionalFieldArray'] = this._toFieldConfig(<JSONSchema7> jsonSchema.additionalItems, key, options);
           }
 
           Object.defineProperty(field, 'fieldArray', {
@@ -68,6 +78,18 @@ export class FormlyJsonschema {
         }
         break;
       }
+    }
+
+    // map in possible formlyConfig options from the widget property
+    if (jsonSchema['widget'] && jsonSchema['widget'].formlyConfig) {
+      const widgetConfig = jsonSchema['widget'].formlyConfig;
+      field = reverseDeepMerge(widgetConfig, field);
+    }
+
+    // if there is a map function passed in, use it to allow the user to
+    // further customize how fields are being mapped
+    if (options && options.map) {
+      field = options.map(field, jsonSchema);
     }
 
     return field;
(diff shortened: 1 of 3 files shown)


src/core/json-schema/src/formly-json-schema.service.ts
```diff
@@ -70,6 +80,18 @@ export class FormlyJsonschema {
       }
     }
 
+    // map in possible formlyConfig options from the widget property
+    if (jsonSchema['widget'] && jsonSchema['widget'].formlyConfig) {
```
aitboudad: I'm asking myself if `widget` is a valid `json-schema` or not :thinking:

src/core/json-schema/src/formly-json-schema.service.ts
```diff
@@ -1,15 +1,25 @@
 import { Injectable } from '@angular/core';
 import { FormlyFieldConfig } from '@ngx-formly/core';
 import { JSONSchema7, JSONSchema7TypeName } from 'json-schema';
+import { ɵreverseDeepMerge as reverseDeepMerge } from '@ngx-formly/core';
```
kenisteward: was the ``` ɵreverseDeepMerge  ``` done like that on purpose? It looks like an extra character got added here somehow.

juristr: Yep that's a "convention" the Angular team uses for not exposing these functions externally, but still allowing them to be shared between libs, hiding them behind this special character 😄. Seems strange but is intended

src/core/json-schema/src/formly-json-schema.service.ts
kenisteward: http://json-schema.org/draft-07/schema

This states you can have any additionalProperties added to a valid jsonSchema json file.  This is basically a jsonSchema that defines jsonSchema syntax :0.  Any who, because of that definition, it is valid to have any fields we want to add to it on it.  The caveat being that those fields won't be parsed by other validators / ui engines unless they explicitly add it.  This will be remedied in draft-08 where we are getting the ability to add "vocabularies" to schemas. That allows implementers to create extensions based on vocabularies and share the vocabulary (e.g. the react json schema guys are working on the UISchema volcabulary).

When all said and done, I do want to make a hyper schema that uses the widget volcabulary so we can have people make valid layout.json with this library without having to go between IDE and browser checking for syntax errors.

I also want to make a companion app that will allow drag/drop layout fields based on the dataSchema for more non technical users to be able to create layout.json but that has nothing to do with your comment haha.

src/core/json-schema/src/formly-json-schema.service.ts
```diff
@@ -70,6 +80,18 @@ export class FormlyJsonschema {
       }
     }
 
+    // map in possible formlyConfig options from the widget property
+    if (jsonSchema['widget'] && jsonSchema['widget'].formlyConfig) {
+      const widgetConfig = jsonSchema['widget'].formlyConfig;
+      field = reverseDeepMerge(field, widgetConfig);
```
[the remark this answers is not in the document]
juristr: ah ok, gonna try

src/core/json-schema/src/formly-json-schema.service.ts
juristr: Yes me too. I've seen it in various places being used also by others (as mentioned in our discussion issue). But it seemed strange as it wasn't included as a valid property in the typings file..

kenisteward: Oh, this is also why I wanted to make sure we had the split concept of "dataSchema" vs "layoutSchema". the data schema should define available fields while the layoutSchema defines what the layout should look like.  We can merge and things behind the scenes still for ease of processing of the fields.  We can also define a  "combinedSchema" concept that still allows for users to just provide that alone.

src/core/json-schema/src/formly-json-schema.service.ts
kenisteward: Thanks!  I learned something new!!

src/core/json-schema/src/formly-json-schema.service.ts
juristr: > This states you can have any additionalProperties added to a valid jsonSchema json file. This is basically a jsonSchema that defines jsonSchema syntax :0. Any who, because of that definition, it is valid to have any fields we want to add to it on it.

@kenisteward if I got what you mean 😅  then that means it's fine to add the `widget` (which also confirms why I've seen it in so many other places

@kenisteward I don't quite get your last comment though. I mean, sure, I always agree to have separation of concerns, thus having two different schemas, for data & UI. That would be the JSONSchema and the UISchema, right?
I just thought we agreed to go with the `widget` so far that somehow is the "layoutSchema" integrated into the JSONSchema for now: https://github.com/ngx-formly/ngx-formly/issues/1459#issuecomment-470631065

That way it's easier to use for ppl 😄. Actually I'm using this PR in my project for experimentation purposes and it's really easy that way to use JSONSchema to create forms. Also the impact on ngx-formly is rather small for now

src/core/json-schema/src/formly-json-schema.service.ts
juristr: sure 👍. You can see them being used for example [here in the Angular repo](https://github.com/angular/angular/blob/master/packages/core/src/core_private_export.ts) :)

src/core/json-schema/src/formly-json-schema.service.ts
aitboudad: in the end, we'll provide both approaches, IMO the `map` callback solved almost everything, 
 WDYT to consider the `widget` as an addon which can be imported by end users in order to use it.
```ts
import { FormlyJsonschema, widgetAddon  } from '@ngx-formly/core/json-schema';

...
fields: FormlyFieldConfig[] = [
  this.formlyJsonschema.toFieldConfig(
    jsonSchema,
    { map: widgetAddon }
  )
]
```

aitboudad: well just leave as it, the current approach seems good anyway.

kenisteward: @juristr 
>@kenisteward if I got what you mean 😅 then that means it's fine to add the widget (which also confirms why I've seen it in so many other places

Yup, that's right!

>I just thought we agreed to go with the widget so far that somehow is the "layoutSchema" integrated into the JSONSchema for now: #1459 (comment)

Doing it this way doesn't allow someone who wants to separate their dataSchema from their uiSchema and share the dataSchema with the backend for validation.  Technically, syntactically you can do this and the backend parser will ignore all widget fields but architecturally if one wanted to have something that just represented the data then they couldn't do that.  They would have to write a mapper that has keyValue pairs of property key to layout field when all they would need is a configuration.

To refer to the example @aitboudad made, I would consider that a JSONSchema that represents how the ui renderings not a JSONSchema that represents how the data should be submitted.  Because of some specifics of how it gets translated in ngx-formly, it did fit the example of the model data that got spit out but there was no representation of the base schema I gave.

@aitboudad 
> WDYT to consider the `widget` as an addon which can be imported by end users in order to use it.

This makes sense if we want to say only support either just the dataSchema (default way we have it now) or dataSchema merged with widget (which you have it for your addon) but I don't think it would be for I have a JSONSchema that represents the data and I have a separate JSONSchema that represents the UI Layout.  Or would you want to do something like have a  "dualSchemaAddon"  that allows you to pass 2 different things to represent each part?

src/core/json-schema/src/formly-json-schema.service.ts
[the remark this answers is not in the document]
juristr: @aitboudad Ok, inverted the order and so it gets overwritten. force-pushed back back up onto the repo.

aitboudad:
Thank you @juristr!

PR #1509, opened 2019-04-02
perf(core): improve exec time of utils

aitboudad:


src/core/src/lib/components/formly.attributes.ts
```diff
@@ -135,11 +132,18 @@ export class FormlyAttributes implements OnChanges, DoCheck, OnDestroy {
   }
 
   private attachAttrElement() {
-    this.fieldAttrElements = [...this.fieldAttrElements, this.elementRef.nativeElement];
+    if (!this.field['_attrElements']) {
```
mohammedzamakhan: instead of using `!` here, just reverse your if else statement

src/core/src/lib/components/formly.field.ts
```diff
@@ -137,7 +137,11 @@ export class FormlyField implements OnInit, OnChanges, DoCheck, AfterContentInit
   }
 
   private attachComponentRef<T extends FieldType>(ref: ComponentRef<T>, field: FormlyFieldConfigCache) {
+    if (!field._componentRefs) {
```
mohammedzamakhan: same here

src/core/src/lib/components/formly.attributes.ts
aitboudad: I don't see the diff, for readability purpose I would prefer my approach (init first and push after)

src/core/src/lib/components/formly.field.ts
```diff
@@ -137,7 +137,11 @@ export class FormlyField implements OnInit, OnChanges, DoCheck, AfterContentInit
   }
 
   private attachComponentRef<T extends FieldType>(ref: ComponentRef<T>, field: FormlyFieldConfigCache) {
```
thorgod: Just a question never seen the format <T extends FieldType>(ref: ComponentRef<T>, field: FormlyFieldConfigCache)
Just wondering why you could not just 
private attachComponentRef <FieldType> (ref: ComponentRef<FieldType>, field: FormlyFieldConfigCache)

aitboudad: just to ensure that the passed ref should extend `FieldType`, in our case 
we pass `FieldWrapper` to attachComponentRef which extend FieldType.

src/core/src/lib/utils.ts
```diff
@@ -10,63 +10,50 @@ export function getFieldId(formId: string, field: FormlyFieldConfig, index: stri
...
-    }
-    field._keyPath = { key: field.key, path };
+    const key = field.key.indexOf('[') === -1
+      ? field.key
+      : field.key.replace(/\[(\w+)\]/g, '.$1');
+
+    field._keyPath = { key: field.key, path: key.indexOf('.') !== -1 ? key.split('.') : [key] };
   }
 
   return field._keyPath.path.slice(0);
```
thorgod: Why do you keep this method in a utils rather then put it directly on the field object?

aitboudad: make sense,
it was used in several places, not only field.

src/core/src/lib/components/formly.attributes.ts
mohammedzamakhan: there is no difference, the only thing is you are doing an extra operation, even though the operation is not expensive, but when it can be avoided just by simply changing the order of if else, then why should we add extra operation

aitboudad: maybe I'm wrong but still don't see the extra operation, `if-else` is one condition similar to `?:` operator, anyway I'll make those changes in my next PR to make everyone happy, thanks!

PR #1511, opened 2019-04-03
fix(core): improve performance for FieldArrayType

### LINKED ISSUE #1489: Large Data Performance FieldArrayType
```
[ ] feature request
[X] support request
```

**Current behavior**
Slow loading lots of fields/data with FieldArrayType
~72 Field Columns * ~36 Rows takes around 20 seconds. 

**Expected behavior**
Faster loading time on large data.

(This is small example I whipped up might not be the right one, I will work on getting and example exactly like the problem I am seeing in production)
https://stackblitz.com/edit/angular-pbbsxv?file=src%2Fapp%2Fapp.component.html

Here is an example of the loading time in production of a large form. I am working on creating an stackblitz example that represents this. 
image

I will be working on also solving this issue... more information to come.

aitboudad:
fix #1489

### DIFF
--- a/src/core/src/lib/extensions/field-form/field-form.ts
+++ b/src/core/src/lib/extensions/field-form/field-form.ts
@@ -1,12 +1,17 @@
 import { FormlyExtension } from '../../services/formly.config';
 import { FormlyFieldConfigCache } from '../../components/formly.field.config';
-import { AbstractControl, FormGroup, FormArray, FormControl, AbstractControlOptions } from '@angular/forms';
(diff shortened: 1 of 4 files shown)


thorgod:
I will test this out.

Would you like any results from it? Like if I encounter errors? Or speed?

aitboudad:
this replaces my previous fix https://github.com/aitboudad/ngx-formly/commit/6164dd4 to avoid relying on angular internal api.

PR #1532, opened 2019-04-16
fix(material): in number input with value 0 label doesn't float. Closes #1029

### LINKED ISSUE #1029: mat-select is empty if selected option has value 0
I have this problem:
image
The mat-select's selected option is "Active", and its value is 0 (NOT "0").
The placeholder (status) doesn't float. I used Augury to check the component's state and I noticed this:
image

I checked why the FormlyWrapperFormField is empty and found the following line of code:
https://github.com/formly-js/ngx-formly/blob/b56c8aa9381215103b194c042e008134da617bd6/src/material/form-field/src/field.type.ts#L63

this.formControl.value in this case is 0 so the expression returns true. Should you check against null or undefined?

Etchelon:
It's a bugfix to avoid that inputs of type number, and also mat-select where mat-options have a [value] that is a number, have their label NOT floating when the bound value is 0, thus resulting in the getter property "empty" to return true.
See #1029 

Before:
image

After:
image

The same code for the  empty getter is in form-field.wrapper.ts but changing only that doesn't seem to have any effect. Maybe you will know whether to edit that piece of code too.

### DIFF
--- a/src/ui-material/src/types/field.ts
+++ b/src/ui-material/src/types/field.ts
@@ -64,7 +64,7 @@ export abstract class FieldType extends CoreFieldType implements OnInit, AfterVi
   get shouldPlaceholderFloat() { return !!this.to.placeholder; }
   get value() { return this.formControl.value; }
   get ngControl() { return this.formControl as any; }
-  get empty() { return !this.formControl.value; }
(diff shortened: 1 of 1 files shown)


aitboudad:
I've tried to adjust your commit without success so I'm going to close in favor of #1538.

Thanks for the PR anyway :+1:

PR #1538, opened 2019-04-17
fix(material): adjust empty value for number/bool type

aitboudad:


src/material/form-field/src/field.type.ts
```diff
@@ -79,7 +79,7 @@ export abstract class FieldType<F extends FormlyFieldConfig = FormlyFieldConfig>
   get value() { return this.formControl.value; }
   set value(value) { this.formControl.patchValue(value); }
   get ngControl() { return this.formControl as any; }
-  get empty() { return !this.formControl.value; }
+  get empty() { return this.value === undefined || this.value === null || this.value === ''; }
```
bhaidar: @aitboudad Where is empty() being called? I tried searching the code but with no luck. Thanks

aitboudad: those properties are only useful when using custom form field control 
 https://material.angular.io/guide/creating-a-custom-form-field-control.

PR #1554, opened 2019-04-25
Adding Material prefix option for the datepicker icon.

ryanjerskine:
This change allows the datepicker icon to be used as a prefix instead of a suffix icon by using to.datepickerOptions.prefixIcon.

The icon is always a suffix icon. Moving it to the left requires brittle css.

The icon can now be left/right aligned without css hacks.

### DIFF
--- a/src/material/datepicker/src/datepicker.type.ts
+++ b/src/material/datepicker/src/datepicker.type.ts
@@ -18,10 +18,10 @@ import { MatDatepickerInput } from '@angular/material/datepicker';
       [placeholder]="to.placeholder"
       [tabindex]="to.tabindex || 0"
       [readonly]="to.readonly">
-    <ng-template #matPrefix *ngIf="to.datepickerOptions.prefixIcon">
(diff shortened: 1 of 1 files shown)


src/material/datepicker/src/datepicker.type.ts
```diff
@@ -18,7 +18,10 @@ import { MatDatepickerInput } from '@angular/material/datepicker';
       [placeholder]="to.placeholder"
       [tabindex]="to.tabindex || 0"
       [readonly]="to.readonly">
-    <ng-template #matSuffix>
+    <ng-template #matPrefix *ngIf="to.datepickerOptions.prefixIcon">
```
aitboudad: what's about `datepickerTogglePosition` that allow the following values:
- suffix: `datepicker-toggle` as a suffix
- prefix: `datepicker-toggle` as a prefix
- null: without datepicker-toggle

ryanjerskine: Sounds good to me. I'll update the PR.

PR #1560, opened 2019-04-26
feat(material): adds datepicker toggle positioning support

ryanjerskine:
Add support for datepickerTogglePosition which support prefix/suffix. By default, the icon is missing.

The toggle currently only supports suffix.

Support for prefix/suffix/none.


ryanjerskine:
Looks much better than my solution. Thanks!

aitboudad:
Thank you @ryanjerskine, it should be available in `V5.1`.

use the watch, to get notified on released!
Selection_005

PR #1655, opened 2019-07-08
eat(json-schema): add support for multipleOf

aitboudad:


aitboudad:
`templateoptuons.min` is an inclusive min, that's why we add `+1` to the `exclusiveMinimum` if passed:
```ts
field.templateOptions.min = schema.exclusiveMinimum + 1;
```

is that correct? or do you think that it should be handled separately?

aitboudad:
removed exclusive for now, it should be done in another PR.

kenisteward:
I think as long as the validation messages match up and are clear we can implement it as such.  
It was just the thought of whether or not we need to let the user know it is an exclusive minimum or not.
Based on the spec 
https://json-schema.org/understanding-json-schema/reference/numeric.html

in draft 4 exclusiveMin/Max were booleans that told you whether min/max should be treated as such.  I don't know if we want to worry about covering old drafts but thought I'd put that here too.

kenisteward:
@aitboudad okie doke sounds good.

aitboudad:
> It was just the thought of whether or not we need to let the user know it is an exclusive minimum or not.

let the user know would be better.

>  I don't know if we want to worry about covering old drafts but thought I'd put that here too.

not sure :thinking: except if someone asked for it.

PR #1658, opened 2019-07-08
feat(json-schema): add support for $ref

aitboudad:


aitboudad:
> I wonder if we should cache all resolved schemas (so a second call doesn't have to make an AJAX request) and allow the use to opt out of caching.

just keep it simple for the moment, a simple async resolver would be enough.

PR #1669, opened 2019-07-14
docs: add info about ngx-translate

johannesjo:

Not sure if this is the best place and way to inform users about it but the readme is usually the first place I personally tend to look for help on these kind of issues.


aitboudad:
I think it better to add it as an example, similar to https://ngx-formly.github.io/ngx-formly/examples/advanced/i18n

johannesjo:
Closing in favor of #1671

Issue #1719, opened 2019-08-14
better cloning for the binding formvalue -> model


lucienbertin:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request
```

**Current behavior**
since version 5.0, when binding the form.value to the model, you now `clone` the current form value to assign it to the model

https://github.com/ngx-formly/ngx-formly/blob/master/src/core/src/lib/components/formly.form.ts#L83
```ts
  changeModel(event: { key: string, value: any }) {
    assignModelValue(this.model, event.key.split('.'), event.value);
    this.modelChange.emit(clone(this.model));
  }
```
while previous versions would not go through a clone
v4 - https://github.com/ngx-formly/ngx-formly/blob/v4/src/core/src/lib/components/formly.form.ts#L71
```ts
  changeModel(event: { key: string, value: any }) {
    assignModelValue(this.model, event.key, event.value);
    this.modelChange.emit(this.model);
    this.checkExpressionChange();
  }
```

_Now_, i have no beef with going through a clone, but my problem is that when you clone objects, you just copy properties and getters, and not all the methods. as a result, when i use a moment for one of my properties, then after the cloning the object is not a moment anymore and i cant use `.format()` or other methods.

a solution would be to patch the method clone, so that objects that know how to clone themselves do it, instead of you copying each property

https://github.com/ngx-formly/ngx-formly/blob/master/src/core/src/lib/utils.ts#L104
```ts
export function clone(value: any): any {
  if (!isObject(value) || value instanceof RegExp || value instanceof Observable || /* instanceof SafeHtmlImpl */ value.changingThisBreaksApplicationSecurity) {
    return value;
  }

  if (value instanceof AbstractControl) {
    return null;
  }

  if (value instanceof Date) {
    return new Date(value.getTime());
  }

  if (Array.isArray(value)) {
    return value.slice(0).map(v => clone(v));
  }

  //add this
  if (value.clone && value.clone()) {
    return value.clone();
  }

  return Object.keys(value).reduce((newVal, prop) => {
    const propDescriptor = Object.getOwnPropertyDescriptor(value, prop);

    if (propDescriptor.get) {
      Object.defineProperty(newVal, prop, { ...propDescriptor, get: () => clone(value[prop]) });
    } else {
      newVal[prop] = clone(value[prop]);
    }

    return newVal;
  }, {});
}
```
(fork and PR incoming)
[...]

PR #1720, opened 2019-08-14
feat(core): better cloning

lucienbertin:

it's a bugfix/feature, depending how you see it. see #1719 for more info

 see #1719

ensure the `clone` method creates an object of the same `type` when this type knows how to clone itself

for example, for a class Foo that implements a method `clone`
```ts
import { clone } from 'utils';
class Foo {
  constructor(public name: string) {}
  clone() { return new Foo(this.name); }
  isFoo() { return true; }
}
const foo = new Foo('test');
const clonedFoo = clone(foo);
console.log(clonedFoo.name); // 'test'
console.log(clonedFoo.isFoo()); // true
```

but for a class that doesn't, the current behavior is unchanged, the clone method copies all properties but not the methods and the type of the original value is lost

```ts
import { clone } from 'utils';
class Bar {
  constructor(public name: string) {}
  isBar() { return true; }
}
const bar = new Bar('test');
const clonedBar = clone(bar);
console.log(clonedBar.isBar()); // throws
```

  - i think it does, but i couldn't use `npm run commit` command
  - see remark

i have issue with the `npm run build`, it fails during ` node .config/build.js` because it needs `cp`

```
> @ngx-formly/common@5.3.0 copy:schematics D:\sites\ngx-formly
> cpr src/schematics dist/@ngx-formly/schematics --delete-first --filter node_modules/
'cp' is not recognized as an internal or external command,
```
and `cp` is not recognize on my win 10 machine.

it comes from [l19 of .config/build.js](https://github.com/ngx-formly/ngx-formly/blob/v5/.config/build.js#L19)
```js
  execSync(`cp README.md ${packagePath}`);
```

not applicable

**Other information**:

i didnt use the one-liner for cloning cuz it did a shallow copy and that broke a test, also it didnt define enumerzable getters, which broke an other test

### DIFF
--- a/src/core/src/lib/utils.ts
+++ b/src/core/src/lib/utils.ts
@@ -118,7 +118,7 @@ export function clone(value: any): any {
     return value.slice(0).map(v => clone(v));
   }
 
-  if (calue.clone && value.clone()) {
+  if (value.clone && value.clone()) {
     return value.clone();
   }
 
--- a/src/core/src/lib/utils.ts
+++ b/src/core/src/lib/utils.ts
@@ -118,21 +118,9 @@ export function clone(value: any): any {
     return value.slice(0).map(v => clone(v));
   }
 
-  if (value.clone && value.clone()) {
-    return value.clone();
-  }
-
-  return Object.keys(value).reduce((newVal, prop) => {
-    const propDescriptor = Object.getOwnPropertyDescriptor(value, prop);
-
-    if (propDescriptor.get) {
-      Object.defineProperty(newVal, prop, { ...propDescriptor, get: () => clone(value[prop]) });
-    } else {
-      newVal[prop] = clone(value[prop]);
-    }
-
-    return newVal;
-  }, {});
+  // best way to clone a js object maybe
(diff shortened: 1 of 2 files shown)


src/core/src/lib/utils.ts
```diff
@@ -118,6 +118,10 @@ export function clone(value: any): any {
     return value.slice(0).map(v => clone(v));
   }
 
+  if (value.clone && value.clone()) {
```
aitboudad: `clone` method seems like a workaround, the right solution is to find a better way to clone a class instance.
could try the following solution instead:
```ts
// https://stackoverflow.com/questions/41474986/how-to-clone-a-javascript-es6-class-instance
value = Object.assign(Object.create( Object.getPrototypeOf(value)), value);
```

aitboudad: + add some tests :)

lucienbertin:
@aitboudad added a spec and used your code for clioning, is anything else missing

lucienbertin:
i redid the PR but cleaner in #1725

Issue #1777, opened 2019-09-13
JSON schema form array type Error: this.formControl.removeAt is not a function 


Adam-Michalski:
[x] bug report
[x] support request

**Current behavior**

After `toFieldConfig(schema)` I get this error `Error: this.formControl.removeAt is not a function ` on validated schema. I don't know why `FormControl` for ArrayField is generated as `FormGroup` and not as `FormArray`

**Expected behavior**

I can remove element from array type without errors, `FormControl` for array type is `FormArray`

**Minimal reproduction of the problem with instructions**

Don't worry about leak of inputs in new elements in example on StackBlitz
arrays in [StackBlitz](https://stackblitz.com/edit/angular-y334nn)
Validated schema => assets > json-schema > arrays.json

lkinasiewicz:
Looks like items removal is broken when:
- there is a schema with `items` as an array of schemas
- the array is entirely filled with data from the beginning

lkinasiewicz:
This fix is incorrect because some tests fail but it shows where the problem is.

aitboudad:
released in `5.4.2` version.

PR #1779, opened 2019-09-13
fix(core): fixed removing items of initially full array (#1777)

lkinasiewicz:
Bug fix

Current behavior:
https://github.com/ngx-formly/ngx-formly/issues/1777
When we create a form that contains an array and this array if initially full, its items cannot be removed.

New behavior:
We can remove items of initially full array.

### DIFF
--- a/src/core/src/lib/extensions/field-form/field-form.ts
+++ b/src/core/src/lib/extensions/field-form/field-form.ts
@@ -8,7 +8,7 @@ import { registerControl } from './utils';
 export class FieldFormExtension implements FormlyExtension {
   onPopulate(field: FormlyFieldConfigCache) {
     // TODO: add an option to skip extension
-    if (field.fieldArray) {
+    if (field.fieldArray || field.type === 'array') {
       return;
     }
 


src/core/src/lib/extensions/field-form/field-form.ts
```diff
@@ -8,7 +8,7 @@ import { registerControl } from './utils';
 export class FieldFormExtension implements FormlyExtension {
   onPopulate(field: FormlyFieldConfigCache) {
     // TODO: add an option to skip extension
-    if (field.fieldArray) {
+    if (field.fieldArray || field.type === 'array') {
```
aitboudad: use `field.hasOwnProperty('fieldArray')` instead

aitboudad:
close in favor of #1784.

src/core/src/lib/extensions/field-form/field-form.ts
lkinasiewicz: The problem with this is that:
1. This won't support prototypal inheritance if someone wanted to extend form configuration, which is quite natural in JS.
2. This will work improperly when some one generates configuration with `fieldArray: null` or `fieldArray: undefined`.
3. This will be a breaking change because of the above.

Are you absolutely sure? If so, I think this should be documented.

aitboudad: right 👍, I've got another way to fix the `json-schema` issue, instead of `null` value I used an empty object when `additionalItems` is empty.

PR #1830, opened 2019-10-11
Allow custom-inline for checkbox, radio and multicheckbox

internalsystemerror:
This is a fairly simple change which allows for custom-inline to be added as an option to the checkbox, radio and multicheckbox fields for bootstrap.

See: https://getbootstrap.com/docs/4.3/components/forms/#inline-1


aitboudad:
our latest minor version of V5 has been released so this should be merged in `V6` instead, so if you need that feature soon on your project I would suggest creating a custom type on your side.
and thanks for the PR 👍!

internalsystemerror:
tx. i'm using these as custom components already, i'll look at adding a test too eventually

aitboudad:
close in favor of #2133

PR #1873, opened 2019-10-24
fix(json-schema): use multi-select to handle anyOf schema

aitboudad:


aitboudad:
Thank you @logvinoleg89!

PR #1943, opened 2019-11-19
fix(core): propagate disabled state to subfields

### LINKED ISSUE #1528: Disabled not propagated to added FieldArrayType rows
```
[x] bug report
```

**Current behavior**
templateOptions.disabled works for the rows with an existing model. Adding new rows however do not get this `disabled` propagated.

**Expected behavior**
On changes of the model or FormlyFieldConfig `disabled` should be propagated (again).

https://stackblitz.com/edit/angular-yg1gif-96ytzh?file=src%2Fapp%2Fapp.component.ts

### LINKED ISSUE #1722: Disable gets lost with FieldArrayType when model is loaded
```
[x] bug report
```

**Current behavior**
When the model changes and the number of array items changes the disabled property is reset (false).

https://stackblitz.com/edit/angular-yg1gif-bamsxg?file=src/app/app.component.html

The FieldArraytype is initially set to disable (true). Klick "Load Model", the disable property is now false.

**Motivation**
I wanted to use this field property to use expressionProperties on the array items as described in #1528, but as the property gets reset I can not go this way
```
   prePopulate(field: FormlyFieldConfig) {
       field.fieldArray.fieldGroup.map((f, index) => {
         f.expressionProperties = {
           ...(f.expressionProperties || {}),
           'templateOptions.disabled': 'field.parent.templateOptions.disabled',  // still have to check if this is correct
         };
       });
   }
```

**Related issues**
#1528
Maybe this is the source of the problem which then causes the issue in #1528.

### LINKED ISSUE #1743: Attribute disabled API
```
[x] feature request
```

**Current behavior**
We are running into a show stopper for our going live in two months.
There are several issues when changing disabled flag programmatically.

**Expected behavior**
As most of the changing of the attributes are done after interacting with the backend or otherwise programmatically would it make sense to have a formly API to set these attributes?
The application finds the field in the FieldConfig[] and calls a method.

**Related Issues**
#1528
#1722 (maybe with some callback to the custom type to process attribute changes in the children)

**Your opinion**
What would be the best strategy to fix these issues?

aitboudad:
fix #1528, fix #1722, fix #1743

### DIFF
--- a/src/core/src/lib/extensions/core/core.ts
+++ b/src/core/src/lib/extensions/core/core.ts
@@ -63,6 +63,7 @@ export class CoreExtension implements FormlyExtension {
         label: '',
         placeholder: '',
         focus: false,
+        disabled: false,
       },
     });
 
--- a/src/core/src/lib/extensions/field-form/field-form.ts
+++ b/src/core/src/lib/extensions/field-form/field-form.ts
@@ -55,7 +55,13 @@ export class FieldFormExtension implements FormlyExtension {
         _validators: validators,
         _asyncValidators: asyncValidators,
         formControl: control,
+        templateOptions: { disabled },
       } = field;
+
+      if (disabled && control.enabled) {
+        control.disable({ emitEvent: false, onlySelf: true });
+        updateValidity = true;
+      }
 
       if (validators !== control.validator) {
         control.setValidators(validators);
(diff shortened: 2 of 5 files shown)


FritzHerbers:
@aitboudad 
Would love to test the behaviour of the fix.
Are you ready to make a release.

PR #1506, opened 2019-04-01
feat(ng-zorro-antd):ng-zorro-antd UI

sxlwar:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
feature: add ng-zorro-antd ui;

**What is the current behavior? (You can also link to an open issue here)**
none

**What is the new behavior (if this is a feature change)?**
Add ng-zorro-antd UI ,implements input, input-number, checkbox, radio,select and textarea, supports all of the features of theses controls.

**Please check if the PR fulfills these requirements**
- [x] The commit messages follow our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ ] A unit test has been written for this change.
- [x] Running `npm run build` produced a successful build. (Unit testing can be done by running `npm test`;)
- [x] My code has been linted. (`npm run lint` to do this. `npm run build` will fail if there are files not linted.)

**Please provide a screenshot of this feature before and after your code changes, if applicable.**
WX20190401-231012
WX20190401-231620

**Other information**:

Keep all params as the same as [ng-zorro-antd](https://ng.ant.design/docs/introduce/en) used. Use it in 'templateOptions' .

```ts
    ...
    key: 'Input',
    type: 'input',
    tempalteOptions: {
         label: 'Input',
         nzSize: 'large'  // Because of nz-input directive defined a input property named nzSize
    }
   ...
```

Exception:  **nzSpan**  nzLabelSpan is used to set the nzSpan attribute to the nz-form-label，as the same, nzControlSpan is for nz-form-control.

PR #1455, opened 2019-03-04
feat(ng-zorro-antd): add ui-ng-zorro-antd

wenqi73:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
feature, ref #1365 

**What is the current behavior? (You can also link to an open issue here)**

**What is the new behavior (if this is a feature change)?**
add ng-zorro-antd ui

**Please check if the PR fulfills these requirements**
- [x] The commit messages follow our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ ] A unit test has been written for this change.
- [x] Running `npm run build` produced a successful build. (Unit testing can be done by running `npm test`;)
- [x] My code has been linted. (`npm run lint` to do this. `npm run build` will fail if there are files not linted.)

**Please provide a screenshot of this feature before and after your code changes, if applicable.**

**Other information**:

PR #1129, opened 2018-08-15
WIP: feat(ng-zorro-antd): initial implementation of NG-ZORRO AntD form types

beeman:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**

Feature.

**What is the current behavior? (You can also link to an open issue here)**

N/A.

**What is the new behavior (if this is a feature change)?**

This PR aims to implement support for the [Ant Design UI Language for Angular](https://ng.ant.design/docs/introduce/en).

**Please check if the PR fulfills these requirements**
- [X] The commit messages follow our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [ ] A unit test has been written for this change.
- [X] Running `npm run build` produced a successful build. (Unit testing can be done by running `npm test`;)
- [X] My code has been linted. (`npm run lint` to do this. `npm run build` will fail if there are files not linted.)

**Please provide a screenshot of this feature before and after your code changes, if applicable.**

**Other information**:

Marked as WIP as I did not try working with it. It probably needs some fixes to make the fields behave as expected.

PR #1951, opened 2019-11-20
feat: add ng-zorro-antd

aitboudad:
fixes #1506, #1455 and #1129


zzhimin:
you can also use this library（More diverse configuration options）：https://github.com/zzhimin/ngx-ng-zorro-formly

PR #1980, opened 2019-12-15
feat(core): support validations for fieldGroup with empty key

### LINKED ISSUE #1476: JSON powered with validation
```
[ ] feature request
[x] support request
```

**Current behavior**
Attempting to have a JSON powered form with validation

**Expected behavior**
Be able to load form field and defined validation from JSON

Refer to this simplified example:
https://stackblitz.com/edit/angular-oqnqnw
I have added a validator to the first name +last name fields to prevent entry of John Doe.
In reality I will have 30+ fields and need to have validation between various fields on the form.

I'm assuming the way to do this is by then have a fieldGroup for those fields (in this case the whole form) so that i can access the appropriate field values (if there's a better way to do this happy to hear about it)
In the example the nameVal function will do the validation of the field values and provide error feedback.

The issue is due to adding the fieldGroup to the form fields - this appears to prevent the JSON data from being loaded into the fields. If i maps the fields to the fieldGroup it loads the data but the event then doesn't fire to perform validation.

`
```
// This is required to perform validation but prevents data loading from json
      this.fields = fields.map(f => {
        if (f.templateOptions && f.templateOptions.change) {
          f.templateOptions.change = Function('$event', f.templateOptions.change).bind(this);
        }
return f;
      });`
```
```

      // This is required to load the data but then validation isn't called

      this.fields = fields[0].fieldGroup;
```

I'd like to have JSON powered forms with some complex validation between multiple fields on a form.

Thanks.

aitboudad:
fix #1476

### DIFF
--- a/demo/src/app/examples/validation/matching-two-fields/app.component.ts
+++ b/demo/src/app/examples/validation/matching-two-fields/app.component.ts
@@ -13,7 +13,6 @@ export class AppComponent {
 
   fields: FormlyFieldConfig[] = [
     {
-      key: 'password',
       validators: {
         fieldMatch: {
           expression: control => {
--- a/src/core/src/lib/extensions/field-form/field-form.spec.ts
+++ b/src/core/src/lib/extensions/field-form/field-form.spec.ts
@@ -144,8 +144,23 @@ describe('FieldFormExtension', () => {
       form: new FormGroup({ test: formControl }),
     });
 
-    expect(formControl.setValidators).toHaveBeenCalledWith(field._validators);
+    expect(formControl.setValidators).toHaveBeenCalledTimes(1);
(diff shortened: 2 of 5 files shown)


kenisteward:
Thank you for this pr!!

AparnaNallamalli:
lifecycle: {
        onInit: (form, field) => {
          
          field.formControl.valueChanges
        
          .pipe(
            takeUntil(this.onDestroy$),
            tap(() => {
              
I need only for formcontrol value but it is taking valuechanges, please help me how to resolve this

aitboudad:
@AparnaNallamalli to get the form control value:

```ts
field.formControl.value
```

aitboudad:
**Note**: This feature has been ported to [v5.7](https://github.com/ngx-formly/ngx-formly/releases/tag/v5.7.0).

PR #2010, opened 2020-01-03
fix(core): render template type when used in expressionProperties

aitboudad:


rogercoder:
This broke my use of ngx-formly. In version 4 I had been emitting field json like this:
`"type":"select",
"template":null,`
This worked fine, because in line 74 the check was for:
`if (field.template) ...`
So if the template were null, everything was fine.
But in this commit the check was changed to:
`if (field.hasOwnProperty('template') ...)`
which is true even if the value of "template" is null.
So now I'm getting the error:
`NgxFormly: passing 'type' property is not allowed when 'template' is set.`
In my opinion the error is not really correct ... if "template" is null it is not actually being "set" to anything.

Is there a reason you are no longer checking to see if the template has been set?


I don't know about the larger issue you were trying to fix, but if you have no objection I could do a pull request with line 77 changed to either:
`(field.hasOwnProperty('template') && field.template)`

rogercoder:
I opened issue #2121

PR #2034, opened 2020-01-15
fix(material): allow overriding tabindex on mat-radio-group

### LINKED ISSUE #2032: tabindex on mat-radio-group is always set to -1
Is there a way to set tabindex in a mat-radio-group?

I tried to use tabindex property in templateOptions in my formly configuration but when rendered is always set to -1

aitboudad:
fix #2032

### DIFF
--- a/src/material/radio/src/radio.type.ts
+++ b/src/material/radio/src/radio.type.ts
@@ -8,7 +8,7 @@ import { MatRadioGroup } from '@angular/material/radio';
     <mat-radio-group
       [formControl]="formControl"
       [formlyAttributes]="field"
-      [tabindex]="-1">
+      [tabindex]="to.tabindex">
       <mat-radio-button *ngFor="let option of to.options | formlySelectOptions:field | async; let i = index;"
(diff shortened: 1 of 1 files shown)


madmiki80:
Thanks a lot, fast resolved :)
Only one question, i am newby on github, how to add this fix to my formly libraries?

Have a nice day

aitboudad:
> Only one question, i am newby on github, how to add this fix to my formly libraries?

It'll be part of the next release fix `5.5.10` (probably Monday), you'll be notified once released.

madmiki80:
Ty solved this but still have a problem. 
Is it possible to set tabindex also in radio buttons contained in radiogroup? 
I always have tabindex set to -1  there.

Ty :)

aitboudad:
material explicitly set to `-1` for radio buttons and there is no way to override it https://github.com/angular/components/blob/master/src/material/radio/radio.ts#L349

madmiki80:
Ok. Thanks a lot for your support and fast response.

Have a nice day!

Issue #1982, opened 2019-12-16
JSON Schema Field Order


johnRambo2k14:
**Question**
Is it possible to render fields in a specific order specified by a JSON schema? right now my API returns an object and its properties appear in a different order every-time I query it.

Is it possible to provide ngx-formly a JSON schema property to render those fields in a specified order?

 example:
```javascript
{
    "title": "Display Types",
    "type": "array",
    "items": {
        "type": "object",
        "required": [
            "id",
            "display"
        ],
        "order": ["id", "display"], //<--- Is it possible or already a feature to do something like this?
        "properties": {
            "display": {
                "description": "what is the display type?",
                "type": "string",
                "enum": [
                    "GRID",
                    "FORM"
                ]
            },
            "id": {
                "description": "Name of display type",
                "type": "string"
            }
        }
    }
}
```

aitboudad:
use `map` options to adjust the generated FormlyFieldConfig:

```ts
function orderSchemaFields(mappedField: FormlyFieldConfig, mapSource: JSONSchema7) {
  if (schema.order) {
    ...
  }

  return mappedField;
}

...
this.formlyJsonschema.toFieldConfig(schema, { map: orderSchemaFields });
```

aghosh-evertz:
Sorry, I didn't get the second solution (map) for maintaining schema properties order.

aghosh-evertz:
Somehow I am still not able to maintain the order in Formly, any guidance would be much appreciated

PR #2056, opened 2020-01-29
feat(json-schema): adding types

cndMarcel:

This PR expands the returned types by adding `date` and `time` type formats.

If your JSON Schema defines a property like `{ "type": "string", "format": "time" }`, then Formly returns the type "string" only.
This means that you don't know if the returned property is supposed to be displayed as a regular input field (current standard behaviour), a time field or a date field.

Formly now additionally returns the types `time` or `date` if the JSON Schema provides properties with the formats `time` or `date` respectively.

According to the latest JSON Schema draft an extension of the instance type is possible:

> "Note that JSON Schema vocabularies are free to define their own extended type system."
> https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.4.2.1

**Other information**:

In case this is classified as a breaking change you could add a flag like "extended instance types" that makes it possible to return `string`, `time` and `date` instead of only `string`?

### DIFF
--- a/src/core/json-schema/src/formly-json-schema.service.ts
+++ b/src/core/json-schema/src/formly-json-schema.service.ts
@@ -410,8 +410,10 @@ export class FormlyJsonschema {
     return [deps, schemaDeps];
   }
 
-  private guessType(schema: JSONSchema7) {
+  private guessType(schema: JSONSchema7): string {
     const type = schema.type as JSONSchema7TypeName;
+    const format = schema.format;
+
     if (!type && schema.properties) {
       return 'object';
     }
@@ -424,6 +426,10 @@ export class FormlyJsonschema {
       if (type.length === 2 && type.indexOf('null') !== -1) {
         return type[type[0] === 'null' ? 1 : 0];
       }
+    }
+
+    if (type === 'string' && format !== undefined) {
+      return format;
     }
 
     return type;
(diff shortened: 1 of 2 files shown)


src/core/json-schema/src/formly-json-schema.service.ts
```diff
@@ -426,6 +428,10 @@ export class FormlyJsonschema {
       }
     }
 
+    if (type === 'string' && format !== undefined) {
+      return format;
```
aitboudad: We should restrict this to the date & time only and check  if the type is defined else we should fallback to the Native type by setting `type` to the `templateOptions`:

```ts
{
  type: 'string',
  templateOptions: {
    type: 'date', // 'time' | 'datetime',
  }
}
```

aitboudad (changes requested):

aitboudad:
`Note`: Just in case you don't know/ or want this feature asap, a `map` options can be used to adjust the generated `FormlyFieldConfig` https://github.com/ngx-formly/ngx-formly/issues/1982#issuecomment-566048527.

src/core/json-schema/src/formly-json-schema.service.ts
cndMarcel: So would changing the implementation to something like this be sufficient?

```typescript
if (type === 'string' && (format === 'date' || format === 'time')) {
  return format;
}
```

I don't quite understand your point about the `templateOptions`. Should they be changed as well?

indiealexh:
This feature is important to me. Anything I can do to help progress this PR?

aitboudad:
@indiealexh see https://github.com/ngx-formly/ngx-formly/issues/2222#issuecomment-623602210

menosprezzi:
@aitboudad I'll restart this PR applying the since the original author does not have applied the suggested changes, ok?

PR #2062, opened 2020-01-30
fix(core): avoid using includes for IE compatibility

### LINKED ISSUE #2058: IE11 - Radio Button Group - Change/Selection Not Occuring
**Question**
Our company uses this library to dynamically generate forms. We've one of our clients is required to use IE, however one of the issues that's been raised is that radio buttons groups & checkboxes aren't working on forms. I couldn't test via your site as the UI examples fail to render. 

From my testing it looks like the change event isn't being fired and the mat-radio-group doesn't get the value of the mat-radio-button selected along with the class update for the mat-radio-group.

Using the same code from Angular Material (which works on IE11) here:
https://material.angular.io/components/radio/examples

However, as soon as I pass this through formly as a custom FieldType (as type 'radio' fails as well), it fails.

Do you have any suggestions or workarounds for this? 

_unfortunately IE isn't capable of plunkr or stackblitz to share an example_

aitboudad:
fix #2058

### DIFF
--- a/demo/src/index.html
+++ b/demo/src/index.html
@@ -10,6 +10,9 @@
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="icon" type="image/x-icon" href="favicon.ico">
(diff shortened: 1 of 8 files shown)


devinci:
Thank you! 🎉

PR #2107, opened 2020-02-19
fix(core): emit statusChanges on updateValueAndValidity

### LINKED ISSUE #2106: Angular Material - validation message not showing if required set dynamically
First of all, thanks alot for the awesome library. It is easy to configure and flexible to modify as we want.

I have an requirement to set field2 as required based on field1. It works absolutely fine but validation message doesn't show properly. Here is the simple example which I have created,

[stackblitz](https://stackblitz.com/edit/ngx-formly-validation-msg)

Steps to reproduce:

1. Enter something on field1 and it shows asterik (*) on field2 but it doesn't show the message as Field required.
2. I tried clicking on the field2 but still it doesn't show message.

I am not sure if this is something to do with Angular material.

aitboudad:
fix #2106


kritisoft:
I think this issue still persists on async validators .
It is Working on the parent form, but when I use it in the child form, I need to manually update it.

aitboudad:
@kritisoft please fill a new issue with a reproduction example, thanks!

PR #2169, opened 2020-04-02
feat(material): mat-hint alignment support

glenngr:

Adds support for aligning mat-hints to the left or right (start or end).
Adds example on how to use to the docs

Mat-hint is supported on form-fields through the description templateOptions property.
However, specifying the alignment of the hint is not supported.

You can display one start-aligned hint and one end-aligned hint. You can specify one or both.

image

**Other information**:

No logic has been added, thus no unit tests added

### DIFF
--- a/demo/src/app/app.component.ts
+++ b/demo/src/app/app.component.ts
@@ -92,6 +92,7 @@ export class AppComponent {
         { path: '/examples/other/advanced-layout-flex', title: 'Advanced Layout (Flex)' },
         { path: '/examples/other/nested-formly-forms', title: 'Nested Forms (fieldGroup wrapper)' },
         { path: '/examples/other/material-prefix-suffix', title: 'Material Field Prefix/Suffix' },
+        { path: '/examples/other/material-formfield-hint-align', title: 'Material Field Hint Alignment' },
         { path: '/examples/other/hide-fields-with-animations', title: 'Hide Fields with `@angular/animations`' },
         { path: '/examples/other/button', title: 'Button Type' },
         { path: '/examples/other/json-powered', title: 'JSON powered' },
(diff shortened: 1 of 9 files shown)


aitboudad:
@glenngr thanks for the PR, I'll check it when time allows :).

glenngr:
@aitboudad Great, thanks.
The only library code that has changed are the two lines here: https://github.com/ngx-formly/ngx-formly/pull/2169/commits/066080248a144037eea24af768fe5c9b74d80ce8#diff-f9b5c78ee270aed839b443fa98d571de

The other changes are documentation.

glenngr:
any news?

aitboudad:
Hi, sorry for the delay right now the current PR has some limitation that needs some adjustments.
What I'm looking for is a solution similar to the prefix and suffix https://formly.dev/examples/other/material-prefix-suffix

glenngr:
The existing support for `mat-hint` is through the description property in templateOptions. Only strings are supported.
The description property translates to `<mat-hint>` in the `form-field.wrapper.ts`

`<mat-hint>` accepts a content-block, so we can implement it the way you suggest.
But what about backwards compatibility and usability? I think having a solution that only supports a template, like the prefix and suffix solution, is more flexible but it also requires more work to use it.

So the library should either support both string and templates, or only strings (because description only supports string).
I was thinking about creating a component to be used within the material form-field-wrapper that accepts either a template or a string and renders it. It could be used both for prefix,/suffix and for hints.

```
      <mat-hint *ngIf="field._matHintEnd as hintEnd" [id]="null" align="end">
        <formly-string-or-template [content]="hintEnd"></formly-string-or-template>
      </mat-hint>
```

What do you think? You will get what you asked for (template support for hints like in prefix/suffix), and in addition you will get string support for both prefix/suffix and hints.

glenngr:
@aitboudad I have updated my PR to reflect your comments and my ideas. I also rebased on master

aitboudad:
Hey, I'll check your change in the following week, promised :pray:!

aitboudad:
LGTM, the only issue is the [ExpressionChangedAfterItHasBeenCheckedError](https://github.com/angular/components/issues/16209) which we need to find a solution before merging

aitboudad:
I think I'm able to fix it, could you please enable "Allow edits" checkbox of this PR:

pr

glenngr:
I am unable to find that checkbox. Where is it supposed to be?

aitboudad:
close in favor of #2646

PR #2202, opened 2020-04-23
fix(ionic): allow async options for ion-select

mburger81:
there is a bug for ion-select when options are async, wrapping anything in an ng-container should be a workaround for this old bug

https://github.com/ionic-team/ionic/issues/19324


aitboudad:
Thank you @mburger81!

aitboudad:
Released in [v5.6.1](https://github.com/ngx-formly/ngx-formly/releases/tag/v5.6.1) version.

PR #2235, opened 2020-05-08
feat(json-schema): move 'null' type to @formly/core

beeman:
Feature

Null type for json-schema needs to be implemented manually.

Null type for json-schema is added to @formly/core

**Other information**:

I think it makes most sense to add the `null` type to core, but open for other suggestions.

I had to get rid of the `form-field` wrapper, not sure if that's an issues. 

@aitboudad what do you think?

### DIFF
--- a/demo/src/app/examples/advanced/json-schema/app.module.ts
+++ b/demo/src/app/examples/advanced/json-schema/app.module.ts
@@ -75,7 +75,6 @@ export function constValidationMessage(err, field: FormlyFieldConfig) {
         { name: 'const', message: constValidationMessage },
       ],
       types: [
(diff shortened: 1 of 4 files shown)


aitboudad:
Well, I'm not clearly sure about that move :thinking:, `null` in `json-schema` is used to display `title` and `description` an may also render error message for that reason we've used `form-field` wrapper.

I'd prefer to not hurry for `null`, `object` and `array` types. They should the latest part in `json-schema` roadmap.

beeman:
That makes total sense @aitboudad - I'm closing this for now. Let's take another look at it further down the line :)

PR #2337, opened 2020-06-28
fix(schematics): add ng-zorro-antd to ui-theme option

### LINKED ISSUE #2333: I can't add ng-zorro-antd
I wanted to use ng-zorro-antd. So I used 'ng add @ngx-formly/schematics --ui-theme=ng-zorro-antd' to add. But I got the following error:
Cannot parse arguments. See below for the reasons.
    Argument --ui-theme could not be parsed using value "ng-zorro-antd". Valid values are: "material", "bootstrap", "ionic", "nativescript", "primeng", 
"kendo".

aitboudad:
fix #2333

### DIFF
--- a/src/schematics/src/ng-add/index.ts
+++ b/src/schematics/src/ng-add/index.ts
@@ -61,7 +61,7 @@ function addUITheme(options: Schema) {
     const uiTheme = options.uiTheme;
 
     if (uiTheme) {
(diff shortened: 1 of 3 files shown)


zzhimin:
you can also use this library：https://github.com/zzhimin/ngx-ng-zorro-formly

PR #2544, opened 2020-10-17
fix(core): assign validator for field with empty key

aitboudad:


kenisteward:
Thank you for this PR. I couldn't quite figure out where the problem was. I was trying to make a or so you wouldn't have to haha.

aitboudad:
released in [v5.10.4](https://github.com/ngx-formly/ngx-formly/releases/tag/v5.10.4).

PR #2683, opened 2021-01-29
feat(core): add fieldChanges to form

manuelmeister:
New Feature: Add an output `fieldChanges` to `formly-form`

It is only possible to listen for changes to the whole model

Allow listening to individual field changes. I need this to patch the redux store with the changed value.


aitboudad:
Thank you for the PR, I'm not in favor of adding an output for that specific case instead there is another way which will be introduced in `v6` that provides this feature, the `valueChanges` is emitter through the `options.fieldChanges`: https://github.com/ngx-formly/ngx-formly/blob/main/src/core/src/lib/components/formly.field.ts#L230 which you could use.

Anyway, I'll make an alpha version next week so you can start using it!

PR #2716, opened 2021-02-25
feat(core): add v6 migration guide

manuelmeister:
I tried to add a migration guide for v6. I'm not sure if the fieldChanges change is a breaking change. 

@aitboudad if someone relied on #1901 they could mitigate it with extending the formGroup and formArray and setting the defaultValue, correct?


aitboudad:
> if someone relied on #1901 they could mitigate it with extending the formGroup and formArray and setting the defaultValue, correct?

correct and better to mention it in the upgrade!

aitboudad (approved):

aitboudad:
Thank you @manuelmeister!

Issue #2570, opened 2020-10-28
Control resetOnHide for a specific field


Suraj786:
**Description**
Hideexpression is not working with repeat and ngx-datatable control

[demo link](https://stackblitz.com/edit/angular-nbze38-ma3dmg)

**Your Environment**
- Angular version: 8.1.2
- Formly version: 5.10.5

**Additional context**
Even we have nested hideexpression inside repeat control. 
Can you please help me find a solution.

Thanks in advance.

aitboudad:
The passed model value for `fieldArray` and `fieldGroup` is the scoped one, what you're looking for is the root model which can be accessed using `field.parent.model` https://github.com/ngx-formly/ngx-formly/issues/704#issuecomment-364661614

Suraj786:
Thanks for the reply,

Now I have added the `field.parent.model` but if you see when we switch radio button options 2-3 times then datatable's initial values will disappear. 
Even on repeater if you click on add another then nested hideexpression wont work in that case.

updated demo link - [https://stackblitz.com/edit/angular-nbze38-retv6u](https://stackblitz.com/edit/angular-nbze38-retv6u)

can you please have a look at this scenario.

Thanks in advance.

aitboudad:
I think that's expected behavior when `resetFieldOnHide` is enabled as Formly expected a valid initial model value (when the radio button is empty the data table data should be empty).

Suraj786:
Hi @aitboudad 

Can we have feature like `autoClear: false` for control level we don't reset for few selected control when on module we have `resetFieldOnHide: true`.
So on above example datatable can retain its initial values.

Can you please have a look at this scenario.
Thanks in advance.

Suraj786:
Hi @aitboudad 

Can you please help me out on above commented scenario. Sorry for bothering you but I am stuck on that issue.
[Above comment](https://github.com/ngx-formly/ngx-formly/issues/2570#issuecomment-738596546)

Thanks in advance.

aitboudad:
did you try to pass `autoClear: false`, its an internal feature which we may expose later

Suraj786:
Hi @aitboudad 

Yes, I have tried adding `autoClear: false` but its showing below error.
Object literal may only specify known properties, and `autoClear` does not exist in type `FormlyFieldConfig`.

Even after removing `FormlyFieldConfig`  type its not working.

I am using Formly `version: 5.10.5`

Thanks in advance.

aitboudad:
[...]

PR #2721, opened 2021-02-28
feat(core): allow control resetOnHide for a specific field

aitboudad:
BREAKING CHANGE: The `resetFieldOnHide` option is enabled by default
fix: #2570

### DIFF
--- a/UPGRADE-6.0.md
+++ b/UPGRADE-6.0.md
@@ -46,12 +46,34 @@ UPGRADE FROM 5.0 to 6.0
   ]
   ```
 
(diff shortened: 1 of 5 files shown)


orlando1108:
Hi, does that new option is applied recursively inside a group of field configs ?
```
fieldConfig {
  resetOnHide: true,
  fields[]: [
    => if parent is hidden, all fields are reset
  ]
}
```

aitboudad:
> if parent is hidden, all fields are reset

Exactly as described :pray:!

PR #2724, opened 2021-02-28
build: upgrade to angular v11

manuelmeister:


manuelmeister:
I don't understand the error message of the build. I don't think I have the capacity and knowledge to migrate to angular 11.

aitboudad:
Next time then :), I'll try to do it when time allows!

PR #2745, opened 2021-03-10
feat(core): add type of templateOptions accessor "to"

manuelmeister:
Feature Request: Use typing of templateOptions given in generic FormlyFieldConfig _F_ in `FieldType<F>.to`

The type of value accessor `FieldType.to` is always `FormlyTemplateOptions`, even when you provide additional typing.
```ts
export declare abstract class FieldType<F extends FormlyFieldConfig = FormlyFieldConfig> {
    field: F;
    get to(): FormlyTemplateOptions;
}
```

The return type of `to` is a `NonNullable` type of `templateOptions` of the generic field _F_ 
```ts
export declare abstract class FieldType<F extends FormlyFieldConfig = FormlyFieldConfig> {
    field: F;
    get to(): NonNullable<Required<F>['templateOptions']>;
}
```

### DIFF
--- a/src/ui/material/form-field/src/form-field.wrapper.ts
+++ b/src/ui/material/form-field/src/form-field.wrapper.ts
@@ -11,7 +11,7 @@ import {
 } from '@angular/core';
 import {
   ɵdefineHiddenProp as defineHiddenProp,
-  FormlyFieldConfig,
+  FormlyFieldConfig as CoreFieldConfig,
   FormlyTemplateOptions,
   FieldWrapper,
 } from '@ngx-formly/core';
@@ -20,14 +20,17 @@ import { FocusMonitor } from '@angular/cdk/a11y';
 import { FloatLabelType, MatFormFieldAppearance } from '@angular/material/form-field/form-field';
 import { ThemePalette } from '@angular/material/core';
 
-export interface MatFormlyFieldConfig extends FormlyFieldConfig {
+interface MatFormlyFieldConfig extends FormlyFieldConfig {
   _matprefix?: TemplateRef<any>;
   _matsuffix?: TemplateRef<any>;
   _formField?: FormlyWrapperFormField;
+}
+
+export interface FormlyFieldConfig extends CoreFieldConfig {
   templateOptions?: MatFormlyTemplateOptions;
 }
 
-export interface MatFormlyTemplateOptions extends FormlyTemplateOptions {
+interface MatFormlyTemplateOptions extends FormlyTemplateOptions {
   prefix?: TemplateRef<any>;
   suffix?: TemplateRef<any>;
   hideLabel?: boolean;
--- a/src/ui/material/form-field/src/form-field.wrapper.ts
+++ b/src/ui/material/form-field/src/form-field.wrapper.ts
@@ -8,21 +8,36 @@ import {
   AfterViewInit,
   TemplateRef,
   ElementRef,
-  ViewContainerRef,
 } from '@angular/core';
 import {
   ɵdefineHiddenProp as defineHiddenProp,
-  ɵobserve as observe,
   FormlyFieldConfig,
+  FormlyTemplateOptions,
   FieldWrapper,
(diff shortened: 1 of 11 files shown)


src/ui/material/form-field/src/form-field.wrapper.ts
```diff
@@ -8,21 +8,36 @@ import {
...
+import { FloatLabelType, MatFormFieldAppearance } from '@angular/material/form-field/form-field';
+import { ThemePalette } from '@angular/material/core';
 
-interface MatFormlyFieldConfig extends FormlyFieldConfig {
-  _matprefix: TemplateRef<any>;
-  _matsuffix: TemplateRef<any>;
+export interface MatFormlyFieldConfig extends FormlyFieldConfig {
+  _matprefix?: TemplateRef<any>;
+  _matsuffix?: TemplateRef<any>;
   _formField?: FormlyWrapperFormField;
```
aitboudad, line 26: those properties should not be exported

src/ui/material/form-field/src/form-field.wrapper.ts
```diff
@@ -8,21 +8,36 @@ import {
...
+  templateOptions?: MatFormlyTemplateOptions;
+}
+
+export interface MatFormlyTemplateOptions extends FormlyTemplateOptions {
```
aitboudad, line 30: no need to export `MatFormlyFieldConfig`, as its part of `MatFormlyFieldConfig` so it can be used through it: 

```patch
- MatFormlyFieldConfig
+ MatFormlyFieldConfig['templateOptions']
```

src/ui/material/multicheckbox/src/multicheckbox.type.ts
```diff
@@ -1,6 +1,7 @@
 import { Component, ChangeDetectionStrategy, ViewChildren, QueryList } from '@angular/core';
 import { FieldType } from '@ngx-formly/material/form-field';
 import { MatCheckbox } from '@angular/material/checkbox';
+import { FloatLabelType } from '@angular/material/form-field/form-field';
```
aitboudad, line 4: not used

aitboudad (changes requested):
also all mat types should now rely on `MatFormlyFieldConfig`:
```ts
export class FormlySliderTypeComponent<MatFormlyFieldConfig> extends ...
```

src/ui/material/form-field/src/form-field.wrapper.ts
aitboudad: better keep `MatFormlyFieldConfig` internal, and export `FormlyFieldConfig`:

```ts
import { FormlyFieldConfig as CoreFieldConfig } from '@ngx-formly/core';

export interface FormlyFieldConfig extends CoreFieldConfig {
  ...
}
```

manuelmeister:
> also all mat types should now rely on MatFormlyFieldConfig:

They do, because they extend from Material FieldType

aitboudad (approved):

manuelmeister:
Could it make sense to extract some templateOptions (e.g. row, cols, …) in the core into own types?

```ts
export type FormlyFieldConfig =
  | TextareaFieldConfig
  | FormlyFieldConfigBase;

export interface TextareaFieldConfig extends FormlyFieldConfigBase {
  type: 'textarea';
  templateOptions?: FormlyTextareaTemplateOptions;
}

export interface FormlyTextareaTemplateOptions extends FormlyTemplateOptions {
  rows?: number;
  cols?: number;
}
```

aitboudad:
The plan is to have a declaration for each type, so the end-user combines those definitions based on the used theme/type.

manuelmeister:
> The plan is to have a declaration for each type, so the end-user combines those definitions based on the used theme/type.

Do you want to merge this PR, or should I implement this in this PR as well?

aitboudad:
better a separate PR, I'll try to merge this PR today (Just need to check it locally).

aitboudad:
Thank you @manuelmeister!

PR #2746, opened 2021-03-10
feat(core): add additionalWrappers attribute

MaxKless:
Feature

Currently, setting a fields wrappers completely overrides the default wrappers defined in the module. 

This PR introduces an "additionalWrappers" attribute which lets you set wrappers to be added after the default ones.

**Other information**:

### DIFF
--- a/src/core/src/lib/models/fieldconfig.ts
+++ b/src/core/src/lib/models/fieldconfig.ts
@@ -83,6 +83,12 @@ export interface FormlyFieldConfig {
   wrappers?: string[];
 
   /**
+   * It is expected to be the name of defined wrappers.
+   * These wrappers will be added to the default ones after instead of replacing them.
+   */
+  additionalWrappers?: string[];
+
+  /**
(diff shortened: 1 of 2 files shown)


MaxKless:
Hi @aitboudad, this is a small feature that I would love to see implemented. I have had to inject the FormlyConfigService to define additional wrappers multiple times while working with formly and I think having an extra attribute would be useful. 
Let me know what you think and if you're willing to take this feature, I will write tests, docs and an example :)

MaxKless:
any thoughts?

aitboudad:
Hi, thanks for the PR!
I think this feature already exists in `v5` using `templateManipulators` (but deprecated as it can be done using a custom extension):

```ts
{
  templateOptions: {
    templateManipulators: {
      preWrapper: [...],
      postWrapper: [...],
    }
  }
}
```

MaxKless:
All right that makes sense. Thanks :)

PR #2793, opened 2021-04-09
fix(schematics): allow passing --project option to ng-add

### LINKED ISSUE #2791: --project option for multi-project solutions
Maybe i missed it but, how do you exactly select the project to which you wish to add formly in a multi-project folder? 

Angular Material supports a --project option to specify the project for installing material.  I assumed that this parameter pertained to the 'ng add' command ... but I was wrong.  
If this is missing (or i am in the wrong) is there a manual workaround to set it up ?  

Assuming you got two apps in a folder called "store" and "admin" If you try to add formly to "admin" you would write..... 

`ng add @ngx-formly/schematics --ui-theme=material --project=admin` 

Currently this gives  `Unknown option: '--project'`

Any pointers would be appreciated.

aitboudad:
fix #2791

### DIFF
--- a/src/schematics/src/ng-add/schema.json
+++ b/src/schematics/src/ng-add/schema.json
@@ -4,6 +4,13 @@
   "title": "@ngx-formly Shell Options Schema",
   "type": "object",
   "properties": {
+    "project": {
+      "type": "string",
+      "description": "The name of the project.",
+      "$default": {
+        "$source": "projectName"
+      }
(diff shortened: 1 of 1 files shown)


Metrophobe:
@aitboudad now that its fixed. Should this work with ng add? (--project)?

aitboudad:
@Metrophobe yes it should work now

PR #2864, opened 2021-05-26
feat(core):  add helper function to get the field by key

### LINKED ISSUE #2854: add helper function to get the field by key
```patch
- const radioField = field.fieldGroup.find(f => f.key === 'firstname');
+ const radioField = field.get('firstname');
```

nirmpate:
add get method to field type

Closes #2854

https://github.com/ngx-formly/ngx-formly/issues/2854

field components will be able to get child field if there is a fieldGroup present. 

- [ x] The commit messages follow our guidelines: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format
- [x ] A unit test has been written for this change.
- [ x] Running `npm run build` produced a successful build. (Unit testing can be done by running `npm test`;)
- [ x] My code has been linted. (`npm run lint` to do this. `npm run build` will fail if there are files not linted.)

### DIFF
--- a/src/core/src/lib/extensions/core/core.ts
+++ b/src/core/src/lib/extensions/core/core.ts
@@ -1,6 +1,6 @@
 import { ChangeDetectorRef, ComponentRef } from '@angular/core';
 import { FormlyConfig } from '../../services/formly.config';
-import { FormlyFieldConfigCache, FormlyValueChangeEvent, FormlyExtension } from '../../models';
+import { FormlyFieldConfigCache, FormlyValueChangeEvent, FormlyExtension, FormlyFieldConfig } from '../../models';
 import {
   getFieldId,
   assignFieldValue,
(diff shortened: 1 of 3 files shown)


MaxKless:
@nirmpate what do you think of enabling recursive searching with an array of rest parameters instead of just searching for the field key until you find it? This way, the user would have granular control of which nested field they're looking for.

src/core/src/lib/extensions/core/core.ts
```diff
@@ -29,6 +29,32 @@ export class CoreExtension implements FormlyExtension {
       });
     }
 
+    Object.defineProperty(field, 'get', {
+      value: (key) => {
+        let formlyField: FormlyFieldConfig = null;
+
+        const findField = (f: FormlyFieldConfig, keyValue: string | number) => {
```
aitboudad, line 36: move this to [util](https://github.com/ngx-formly/ngx-formly/blob/main/src/core/src/lib/utils.ts) file.

src/core/src/lib/extensions/core/core.ts
```diff
@@ -29,6 +29,32 @@ export class CoreExtension implements FormlyExtension {
...
+          if (f.fieldGroup && f.fieldGroup.length > 0) {
+            const match = f.fieldGroup.find((fg) => fg.key && fg.key === keyValue);
```
aitboudad, line 38: we need to manage two extra cases:
- nested key `address.city`.
- finding field by `id` which is useful for a field that doesn't have a key

aitboudad (changes requested):

nirmpate:
> @nirmpate what do you think of enabling recursive searching with an array of rest parameters instead of just searching for the field key until you find it? This way, the user would have granular control of which nested field they're looking for.

This is an interesting idea.  I'm not exactly sure if I understand it though.  So if the get function can take multiple arguments such as field.get(...args).  This array of args would target a specific path to the field you're searching for?

MaxKless:
> This is an interesting idea. I'm not exactly sure if I understand it though. So if the get function can take multiple arguments such as field.get(...args). This array of args would target a specific path to the field you're searching for?

Yes exactly. Check out the [relevant typescript docs](https://www.typescriptlang.org/docs/handbook/2/functions.html#rest-parameters-and-arguments). I imagine you could just iterate over the array recursively and find the nested keys.

aitboudad:
close in favor #3040

PR #2885, opened 2021-06-10
fix(json-schema): handle array type without items keyword

### LINKED ISSUE #2884: Certain schemas can cause unhandled `undefined` in Formly JSON parser
**Description**

Came to light in discussion of issue #2875. Schemas such as

```json
{
  "type": "object",
  "properties":{
    "prop":{
      "type":"array"
    }
  }
}
```

cause the parser to call `FormlyJsonschema.resolveSchema(schema, options)` with `schema` set to the unexpected value `undefined`. This function is not hardened against such a possibility, and errors out before the schema can be parsed.

**Minimal Reproduction**

https://stackblitz.com/edit/extended-formly-config-does-not-intercede?file=src%2Fapp%2Fapp.component.ts

**Your Environment**

- Angular version: 8.2.14
- Formly version: 5.10.20

camden-penn:

Fix issue in which an array without an items keyword causes an error in
FormlyJsonschema.toFieldConfig().

fix #2884

FormlyJsonschema.toFieldConfig() now returns `{}` for items in an array without an `items` keyword, instead of throwing an error.

### DIFF
--- a/UPGRADE-6.0.md
+++ b/UPGRADE-6.0.md
@@ -87,3 +87,12 @@ UPGRADE FROM 5.0 to 6.0
   + },
   }),
   ```
+
+@ngx-formly/bootstrap
+---------------------
+ * add-on: The two first argument of `onClick` handler has been replaced by `field` instance.
+
+  ```patch
+  - onClick: (to, fieldType, $event) => ...,
+  + onClick: (field, $event) => ...,
+  ```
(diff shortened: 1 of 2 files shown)


camden-penn:
An equivalent patch would most likely also work for the `v5` branch. I had trouble getting `v5` to build and was unsure of the protocol to suggest a patch that would work on multiple branches, so that assertion remains untested at the moment.

aitboudad:
close in favor of #2914

PR #2917, opened 2021-07-09
Ion Select Multiple fix

itcane-hatem:
Bug fix

In case of multiple select, the selected values are not checked when opening the dialog again.

In case of multiple select, the selected values are checked

- [] A unit test has been written for this change.
- [] Running `npm run build` produced a successful build. (Unit testing can be done by running `npm test`;)
- [] My code has been linted. (`npm run lint` to do this. `npm run build` will fail if there are files not linted.)

**Other information**:


src/ui/ionic/select/src/select.type.ts
```diff
@@ -31,6 +31,9 @@ export class FormlyFieldSelect extends FieldType {
     templateOptions: {
       options: [],
       compareWith(o1: any, o2: any) {
+        if (Array.isArray(o2)) {
```
aitboudad, line 34: This should be only applied when `multiple` is set to `true`.

PR #3115, opened 2022-01-01
Support input type (text or tel) for datepicker

frankyvij:
**Improvement:** adding support for input type in datepicker.ts by using existing implementation. This would allow users to set the type to "text" or "tel" based on their requirements. A more common use case is to set the type to "tel" to open the numeric keyboard on mobile devices.

**Current behavior:** there is no way to change the input type (defaulting it to text).


aitboudad:
Thanks for the PR, seems there is an alternative way to achieve the same task (`attributes`):

```patch
fields: FormlyFieldConfig[] = [
  {
    key: 'Datepicker',
    type: 'datepicker',
    templateOptions: {
+     attributes: { type: 'number' }
    },
  },
];
```

PR #3154, opened 2022-01-26
fix(core): do not emit `modelChange` when a field without key changes

aitboudad:


kenisteward:
This is going to be in v6 right? 
I'm assuming that because it would be a breaking change.

aitboudad:
yes `v6` also not a breaking change since it doesn't occurs on `v5`.

kenisteward:
Oh okay thanks! I was going to adjust our internal wrapper unit tests to
double check for this otherwise.

Issue #2728, opened 2021-03-01
showValidation extra, analogue to showError


MaxKless:
**Describe the solution you'd like**
I'm using formly to build complex forms (and loving it!), including some fields that contain a mixture of standard and asynchronous validation. To precisely control when to show validation messages, icons etc. the showError extra is a great feature and allows me to centrally define when to show validation errors. However, in some cases I want to not only control when to show errors but when to show any validation at all and when to leave a field blank. I would love to be able to pass another function to extras that allows me to define this behavior. 

**Describe alternatives you've considered**
I could also define when to show validation in each component where I need to control this. However, this would not allow for central configuration (and especially reconfiguration for in different parts of the app using FormlyModule.forChild() )

**Additional context**
I'm willing to try and implement this feature myself and submit a PR. However, I would like to know whether you would even consider this functionality or if my time would be wasted.

aitboudad:
> I'm willing to try and implement this feature myself and submit a PR

All validators are wrapped within the following function [field-validation.ts#L126](https://github.com/ngx-formly/ngx-formly/blob/main/src/core/src/lib/extensions/field-validation/field-validation.ts#L126) so adding just an extra check would do the job.

But I'm not sure if that would be useful or not so let me think about it for a moment.

MaxKless:
I'm still open to trying this, have you come to a conclusion?

aitboudad:
After re-checking, I see the need of a such feature. This will fix the following issue https://github.com/ngx-formly/ngx-formly/issues/2385 as well.

So I'm open to any good proposal form your side :pray:!

MaxKless:
Okay, I will try to implement this! I'll update you when I get somewhere.

MaxKless:
@aitboudad After reading the other issue again, I don't really see how a `showValidation` extra would fix #2385. Can you explain?

aitboudad:
just go ahead with your solution I may see if it can solve the mentioned issue as well, my thought was with `showValidation` you can have a full control of validation errors including which message to display but not sure if that make sense or not.

Issue #2385, opened 2020-07-23
show multiple error validation messages at the same time


fabioportieri:
Hi,

followup of https://github.com/ngx-formly/ngx-formly/issues/2382

is there a way to show all the registered validators error messages? so far i'm under the impression only one error message at a time get shown, with multiple validators

this is a stackblitz of the usecase, the date component has several failed validations in place, only one get shown

https://stackblitz.com/edit/angular-qgebin?file=src%2Fapp%2Fapp.component.ts

a solution with <formly-validation-message  or similar is very welcomed since it promotes a high grade of customization of said error(s)

thank you very much

wesleydean:
We are currently migrating angular-formly over to ngx-formly and our previous implementation displayed multiple validation messages. Is this on the roadmap?

Thanks

dc-p8:
I might be wrong but I feels like the only code that needs to be updated is https://github.com/ngx-formly/ngx-formly/blob/5f099fddea6426b46a453866adf0fede5be0bcd7/src/core/src/lib/templates/formly.validation-message.ts#L45

That's my current workaround:
```ts
{
    // ...
    validators: {
        multipleErrors: {
            expression: (
                control: FormControl,
                field: FormlyFieldConfig
            ) => {
                const v = control.value as string;
                if (!v) return;
                const lines = v
                    .split(/\r?\n|\r/)
                    .map((line) => line.trim())
                    .filter((line) => line.length > 0);
                const totalCharsCount = lines.reduce(
                    (acc, a) => acc + a.length,
                    0
                );
                const _metaErrors = ((field as any)._metaErrors =
                    {} as any);
                if (totalCharsCount > 40) {
                    _metaErrors.totalCountTooLong = true;
                }
                if (lines.some((line) => line.length > 10)) {
                    _metaErrors.oneLineTooLong = true;
                }
                if (lines.length > 4) {
                    _metaErrors.tooMuchLines = true;
                }
                return Object.keys(_metaErrors).length == 0;
            },
                message: (error: any, field: FormlyFieldConfig) => {
                    const ret = [];
                    const _metaErrors = (field as any)._metaErrors as any;
                    if (_metaErrors.totalCountTooLong)
                        ret.push('40 characters max in total');
                    if (_metaErrors.oneLineTooLong)
                        ret.push('10 characters max per line');
                    if (_metaErrors.tooMuchLines)
                        ret.push('4 lines max');
                    return ret.join(`
                `);
                },
        },
    }
[...]

PR #3171, opened 2022-02-07
feat(core): add showValid configuration extra

MaxKless:
THIS IS A DRAFT PR

This is a feature to add a centrally configurable `showValid` extra that types and wrappers can read to configure their validation display behavior.

Currently, it is possible to configure when to show errors using `showError`. This is sometimes too limited as explained in #2728.

Now, an additional function (`showValid`) is available to separately control when to show valid feedback (positive validation) messages and styling. This is not used throughout formly itself but e.g. bootstraps supports it and it is very useful to have a configurable analogue in the module

@aitboudad this is what I had in mind, it's a pretty small change in the core functionality. The custom type in the provided example is adjusted to show what I mean.
Of course, this is a draft PR so if you still think this is a useful feature (even though it doesn't address your concerns about #2385 ) I will proceed and flesh it out with tests and see how to fix the change detection in the example.
I'm also willing to work on a more general solution and probably more complex solution for more fine-grained control over validation. In that case just reject this PR and we can discuss how to proceed.

**Other information**:

### DIFF
--- a/src/core/src/lib/services/formly.config.ts
+++ b/src/core/src/lib/services/formly.config.ts
@@ -42,6 +42,9 @@ export class FormlyConfig {
         (field.formControl?.touched || field.options.parentForm?.submitted || !!field.field.validation?.show)
       );
     },
+    showValid(field: FieldType) {
+      return field.formControl?.touched || field.options.parentForm?.submitted || !!field.field.validation?.show;
+    },
   };
   extensions: { [name: string]: FormlyExtension } = {};
   presets: { [name: string]: FormlyFieldConfig | FormlyFieldConfigPresetProvider } = {};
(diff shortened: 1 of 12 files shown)


aitboudad:
Thanks for the PR :+1:,  I may see if we can unify the current approach with some related issues else we can go that way!

MaxKless:
Okay great, let me know how I can help :) I'm ready to do some actual coding in here!

src/core/src/lib/services/formly.config.ts
```diff
@@ -35,6 +35,9 @@ export class FormlyConfig {
         (field.formControl?.touched || field.options.parentForm?.submitted || !!field.field.validation?.show)
       );
     },
+    showValid(field: FieldType) {
+      return field.formControl?.touched || field.options.parentForm?.submitted || !!field.field.validation?.show;
```
FritzHerbers, line 39: @MaxKless 
Maybe needs to check `field.formControl.valid`.

MaxKless: Yeah good point, fixed.

PR #3195, opened 2022-02-22
fix(material): prioritize matDatepickerParse error over required

### LINKED ISSUE #3190: Required error is output on top of matDatepickerParse in mat-datepicker.
https://stackblitz.com/edit/ngx-formly-ui-material-zbq7ts?file=src%2Fapp%2Fformly-datepicker.component.ts

- Angular version: 9.1.13/9.2.4
- Formly version: 5.10.32

For some reason, the matDatepickerParse error is displayed on top of the required error, this can be seen in the demo. If you remove required from the config, then the behavior will be expected.(the same if you set required: true) in the datapicker template.(BUT without required in the config).
Moreover, the error matDatepickerMin is displayed normally, for example, enter 11/11/1111 in the demo above.
In your official demo, the same pattern is observed if you enter letters in the field.
https://stackblitz.com/angular/pxgndggqamkq?file=src%2Fapp%2Fapp.component.ts

aitboudad:
fix #3190

### DIFF
--- a/src/material/datepicker/src/datepicker.type.ts
+++ b/src/material/datepicker/src/datepicker.type.ts
@@ -1,7 +1,8 @@
-import { Component, ViewChild, AfterViewInit, TemplateRef } from '@angular/core';
+import { Component, ViewChild, AfterViewInit, TemplateRef, OnDestroy } from '@angular/core';
 import { FieldType } from '@ngx-formly/material/form-field';
 import { MatInput } from '@angular/material/input';
 import { MatDatepickerInput } from '@angular/material/datepicker';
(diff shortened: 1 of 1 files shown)


km4dlpx:
How to handle above issue in the latest version (6.1.8) as Wrapproperty is no longer supported by ngx-formly/core?

km4dlpx:
> How to handle above issue in the latest version (6.1.8) as Wrapproperty is no longer supported by ngx-formly/core?

Fixed the above issue using **observe** instead of wrapproperty (import {ɵobserve as observe} from '@ngx-formly/core', replace 'errors' argument with ['error'] and use setErrors instead of **'__$errors'**.

With setErrors method is getting called infinite time and to avoid that need to add a check _.isEqual(currentValue, previousValue)

PR #3203, opened 2022-02-24
feat(core): introduce priority to sort extensions

MaxKless:
core feature

Currently, there is no way to specifically order extensions. It will happen implicitly, depending on the order in which they are defined. This quickly becomes intransparent when you enforce alphabetic sorting or define extensions throughout multiple modules where the loading order is not clear.

This PR introduces an optional `priority` property to the `FormlyExtension` type that enables you to control this behavior. Lower numbers are processed first, with formly's own extensions all get a value of `-1`. Default `priority` is `1`.

Let me know what you think of this feature and if you like it, I will write documentation, an example and everything to make it complete.

### DIFF
--- a/demo/src/app/guides/custom-formly-extension.md
+++ b/demo/src/app/guides/custom-formly-extension.md
@@ -98,3 +98,22 @@ See live demo: [demo](https://stackblitz.com/edit/ngx-formly-ui-bootstrap-slzm3p
 
   Look at the demo to see that the default label has been automatically added to the field.
 
+## Extension priority
+When registering an extension with the `FormlyModule`, you can provide an additional `priority` property of type `number`.
+This will be used to change the order in which extensions are executed. If you have multiple extensions that change the same properties of a `FormlyFieldConfig`, you can use this to ensure they are executed in the correct order. Extensions with higher `priority` values will be executed later.
+
+If you don't provide a `priority` option, a default value of `1` will be used. <br/>
+Formly's own internal extensions have a priority of `-1`.
+
+```typescript
+  FormlyModule.forRoot({
+    extensions: [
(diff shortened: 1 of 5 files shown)


src/core/src/lib/extensions/core/core.ts
```diff
@@ -17,6 +17,8 @@ import { Subject } from 'rxjs';
 
 /** @experimental */
 export class CoreExtension implements FormlyExtension {
+  priority = -1;
```
aitboudad, line 20: instead I would prefer passing priority through `NgModule` config:
```ts
extensions: [
  { name: 'core', extension: core, priority: ... },
  ...
],
```

MaxKless: Okay good point, I will change it to work like that.

MaxKless:
@aitboudad I changed it to configure 'priority' in the module. 

I'm not sure about `FieldArray`, it's kind of a special case and is not registered in any module so I can't really set its prio to `-1` per default. Should I build a mechanism that allows priority definitions in both `ExtensionOption` and `Extension`?

I also built a bit more complicated sorting logic in `FormlyConfig` so that the arrays don't have to be re-sorted every time.

aitboudad:
@MaxKless Thank you!

aitboudad:
> I'm not sure about `FieldArray`, it's kind of a special case and is not registered in any module so I can't really set its prio to `-1` per default. Should I build a mechanism that allows priority definitions in both `ExtensionOption` and `Extension`?

`FieldArray` is not an extension, for field type the order execution rely on `CoreExtension` which has the highest priority by default so I think the current implementation would be enough.

PR #3205, opened 2022-02-25
build: update build script for Windows

MaxKless:
Because they rely on linux/unix command line tools, the build scripts previously didn't work for windows.

Now, the functionality is reimplemented using Node.js functions that work in any environment.

**Other information**:

### DIFF
--- a/build/schematics.ts
+++ b/build/schematics.ts
@@ -1,6 +1,6 @@
-import { exec } from './util';
+import { copyDirSync, exec } from './util';
 
 // build
 exec('cd src/schematics && npm run build');
 // copy to dist
-exec('cp -r src/schematics dist/@ngx-formly/schematics');
+copyDirSync('src/schematics', 'dist/@ngx-formly/schematics');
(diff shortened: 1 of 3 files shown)


build/schematics.ts
```diff
@@ -1,6 +1,20 @@
 import { exec } from './util';
+import * as fs from 'fs';
+import path = require('path');
```
aitboudad, line 3: use `import * as path from 'path';` instead of `require`

build/schematics.ts
```diff
@@ -1,6 +1,20 @@
...
 
 // build
 exec('cd src/schematics && npm run build');
 // copy to dist
-exec('cp -r src/schematics dist/@ngx-formly/schematics');
+copyDirSync('src/schematics', 'dist/@ngx-formly/schematics');
+
+function copyDirSync(src, dest) {
```
aitboudad, line 10: move `copyDirSync` to `util` file.

aitboudad (changes requested):

build/schematics.ts
MaxKless: weird that it automatically used `require`..- fixing :)

aitboudad:
Thank you @MaxKless!

PR #3218, opened 2022-03-08
feat(core): add support of strictTemplates in FieldType

### LINKED ISSUE #2842: Type 'AbstractControl' is not assignable to type 'FormControl'
**Description**

Using strict template checking prints `Type 'AbstractControl' is not assignable to type 'FormControl'.ngtsc(2322)`, because the FormControlDirective requires a FormControl not AbstractControl. Is there a reason why AbstractControl was used as the control type? I think it would change a lot because we always only push FormControls to an input.

**Minimal Reproduction**

```html
<input matInput [formControl]="formControl" />
```

```json
  "angularCompilerOptions": {
    "enableI18nLegacyMessageIdFormat": false,
    "strictInjectionParameters": true,
    "strictInputAccessModifiers": true,
    "strictTemplates": true
  }
```

**Your Environment**

- Angular version: latest
- Formly version: latest

**Additional context**

aitboudad:
fix #2842


aitboudad:
Sorry to hear that, our intention is to keep backward compatible between minor release so if that happens it should be reverted or fixed. So could you please provide us more info about the raised error in order to check it on our side and provide a fix asap.
Thank you!

rstaylor:
The error message I receive when I attempt to 'ng serve' my project is:
 
Error: node_modules/@intersystems/isc-form/lib/custom-fields/repeat/repeat.component.d.ts:9:5 - error TS2416: Property 'field' in type 'RepeatComponent' is not assignable to the same property in base type 'FieldArrayType<FieldArrayTypeConfig>'.
      Type 'Field' is not assignable to type 'FieldArrayTypeConfig'.
        Types of property 'formControl' are incompatible.
          Type 'AbstractControl' is missing the following properties from type 'FormArray': controls, at, push, insert, and 6 more.
    
    9     field: Field;
          ~~~~~
This is occurs where the component has overridden the Field type.  The pertinent part of the component definintion is:

```
import { Field } from "../../interfaces/isc-form-model";
import * as ɵngcc0 from '@angular/core';
export declare class RepeatComponent extends FieldArrayType implements OnInit {
    subscriptions: Subscription[];
    field: Field;

```
The root cause appears to be in the implementation of FieldArrayType where the definition now looks like:

```
export declare abstract class FieldArrayType<F extends FormlyFieldConfig = FormlyFieldConfig> extends FieldType<F> implements FormlyExtension
```
If I remove the "FormlyFieldConfig =" from the extends within the generic type the error goes away.  

For now I have downgraded to ngx-formly/core@5.10.3 which is allowing my project to compile

aitboudad:
I'm able to reproduce it, I'll let you know, once fixed!

aitboudad:
The above issue has been fixed and released as part of [v5.11.2](https://github.com/ngx-formly/ngx-formly/releases/tag/v5.11.2) release.
Thank you!

rstaylor:
@aitboudad Thanks for the quick response.  You were so fast I didn't notice this last update until today.  I wasn't expecting anything.  Great job!

PR #3242, opened 2022-03-26
feat(core): improve Formly config structure

### LINKED ISSUE #2853: Improve Formly config structure
This is part of #2571, I think we need to improve our Formly config for better readability (avoid long JSON), easy to remember, and quick typing.

1- `templateOptions` => `ui` (or `props` :thinking:):

```ts
{
  key: 'name',
  type: 'input',
  ui: { label: 'Name' },
  expressions: {
    'ui.disabled': 'model.disabled'
  }
}
```

2- `expressionProperties|hideExpression` => `expressions`:

```ts
{
  key: 'name',
  type: 'input',
  expressions: {
    hide: 'model.disabled',
  }
}
```

3- `fieldGroup` => `group`:

```ts
{
  key: 'address',
  group: [ ... ]
}
```

4- `fieldArray` => `array`:

```ts
{
  key: 'addresses',
  array: { ... }
}
```

let me know if you've more suggestions!

aitboudad:
fix #2853


src/core/src/lib/models/fieldconfig.ts
```diff
@@ -34,7 +34,10 @@ export interface FormlyFieldConfig {
   /**
    * This is reserved for the templates. Any template-specific options go in here. Look at your specific template implementation to know the options required for this.
    */
-  templateOptions?: FormlyTemplateOptions;
+  props?: FormlyFieldProps;
```
manuelmeister, line 37:
```ts 
export interface FormlyFieldConfig<Props extends FormlyFieldProps> {
  …
  props?: Props;
```

This could be a generic. Then you could make union types of the fields and get better type autocompletion.

E.g:
```ts
export interface TextareaFieldConfig extends FormFieldConfig<TextareaProps> {
  type: "textarea";
}

export interface TextareaProps extends FormlyFieldProps {
  row: number;
  cols: number;
}

export type FieldConfigTypes = TextareaFieldConfig | TelephoneFieldConfig | …;
```

And then I would urgently remove additionalProps as this makes accurate typing much more difficult.

For the transition I would allow these additionalProps only at `templateOptions` and not at `props`.

src/core/src/lib/templates/field.type.ts
```diff
@@ -37,12 +37,13 @@ export abstract class FieldType<F extends FormlyFieldConfig = FormlyFieldConfig>
     return this.field.formControl as F['formControl'];
   }
 
+  get props() {
+    return (this.field.props || {}) as NonNullable<F['props']>;
```
manuelmeister, line 41:
```suggestion
  get props(): NonNullable<F['props']> {
    this.field.props ??= {};
    return this.field.props;
```

Is this not needed?

manuelmeister (commented):
@aitboudad I tried to look at some changes.

aitboudad:
@manuelmeister would you like provide a PR for these changes?

manuelmeister:
I'm currently quite busy, so I'm don't know when I have time to implement this.

PR #3295, opened 2022-05-14
fix(core): remove additionalProps from FormlyFieldProps interface

aitboudad:
fix #3242

### DIFF
--- a/src/core/src/lib/models/fieldconfig.ts
+++ b/src/core/src/lib/models/fieldconfig.ts
@@ -5,7 +5,7 @@ import { FieldWrapper } from '../templates/field.wrapper';
 import { ValidationMessageOption } from '../models';
 import { Type } from '@angular/core';
 
-export interface FormlyFieldConfig {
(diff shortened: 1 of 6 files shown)


src/core/src/lib/models/fieldconfig.ts
```diff
@@ -5,7 +5,7 @@ import { FieldWrapper } from '../templates/field.wrapper';
 import { ValidationMessageOption } from '../models';
 import { Type } from '@angular/core';
 
-export interface FormlyFieldConfig {
+export interface FormlyFieldConfig<Props = FormlyFieldProps & { [additionalProperties: string]: any }> {
```
manuelmeister, line 8:
```suggestion
export interface FormlyFieldConfig<Props = FormlyFieldProps> {
```

Why does this again have the additionalProperties? I thought this should only be allowed on templateOptions.

aitboudad: `FormlyFieldConfig` should allow additionalProperties, to make it strict simply use `FormlyFieldConfig<FormlyFieldProps>`. You may check #3303 to have a clear idea for `strict` mode.

PR #3301, opened 2022-05-19
perf(core): avoid rebuild all fields on FieldArray update

aitboudad:


aitboudad:
@jma please check https://github.com/ngx-formly/ngx-formly/releases/tag/v5.12.2 which improve the performance issue in array type.

PR #3306, opened 2022-05-22
docs(preset): add preset to visible files in example

MaxKless:
When I opened #3256 , there was a mistake in the example - salutation.preset.ts wasn't rendered.

I added it to the config module.


aitboudad:
Thank you @MaxKless!

PR #3342, opened 2022-06-08
fix(core): fixed FieldArrayTypeConfig props declaration

### LINKED ISSUE #3341: v6 FieldArrayTypeConfig does not have generic prop value
**Description**
The FieldArrayTypeConfig type does not allow for a custom props value like FormlyFieldConfig does.

**Minimal Reproduction**
https://github.com/ngx-formly/ngx-formly/blob/a44f2a50a581acdada4bab4803a755a857fa866c/src/core/src/lib/templates/field-array.type.ts#L8

**Your Environment**

- Angular version: 13.3.0
- Formly version: 6.0.0-beta.2

**Additional context**
My guess is this is an oversight, as I don't see why it couldn't be the same as FieldTypeConfig or FieldGroupTypeConfig.

https://github.com/ngx-formly/ngx-formly/blob/a44f2a50a581acdada4bab4803a755a857fa866c/src/core/src/lib/templates/field.type.ts#L5

I'm in the process of updating from 6.0.0-next.9 and taking advantage of the changes, but noticed this type wasn't updated so I have to extend FieldArrayTypeConfig instead like I have been (posted for reference), instead of using props.

https://github.com/dereekb/dbx-components/blob/4b0699c411f04d08982947f981993a87edbe2b46/packages/dbx-form/src/lib/formly/field/value/array/array.field.component.ts#L12

dereekb:
Fixes #3341. 

The FieldArrayTypeConfig type does not allow for a custom props value like FormlyFieldConfig does.

The FieldArrayTypeConfig type will now allow for a custom props value like FormlyFieldConfig does.

**Other information**:

Looks like the types for this interface were overlooked in #3295.

### DIFF
--- a/src/core/src/lib/templates/field-array.type.ts
+++ b/src/core/src/lib/templates/field-array.type.ts
@@ -5,9 +5,9 @@ import { clone, assignFieldValue, getFieldValue, hasKey } from '../utils';
 import { FormlyFieldConfig, FormlyExtension } from '../models';
 import { registerControl, unregisterControl, findControl } from '../extensions/field-form/utils';
 
-export interface FieldArrayTypeConfig extends FormlyFieldConfig {
+export interface FieldArrayTypeConfig<T = FormlyFieldConfig['props']> extends FormlyFieldConfig<T> {
   formControl: FormArray;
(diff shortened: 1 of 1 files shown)


aitboudad:
Thank you @dereekb!

PR #3404, opened 2022-08-02
build: update package-lock.json to node 16

### LINKED ISSUE #3405: Remove problematic "bundlesize" dependency
**Executing `npm i` does not work out of the box for windows users!**

Deep down in the dependency graph, there's the following package: https://www.npmjs.com/package/iltorb 
It requires you to compile the binary yourself so you need to install Python and/or Visual Studio Build Tools... which noone wants, obviously.

This dependency on `iltorb` comes from `bundlesize`, which looks unmaintained. `iltorb` itself is also deprecated.

I would like to replace `bundlesize` somehow so that windows users can check out formly without installing extra stuff.

As a quick fix, you can delete the `bundlesize` dep locally and develop like that.

-------------------

**Additional Info**

image

MaxKless:
After updating Node.js to 16.16.0, the package-lock.json is outdated.

=> Updated

Fix #3405


aitboudad:
Thank you @MaxKless!

PR #3488, opened 2022-11-07
feat(kendo): update peerDependencies to latest Kendo versions

jm-trac-tevans:
Updated kendo versions, including bumping major version for "@progress/kendo-angular-inputs" as major version "breaking" change is only to include additional kendo dependencies so is actually non-breaking, see: https://www.telerik.com/kendo-angular-ui/components/inputs/changelog/


aitboudad:
close in favor of #3592

PR #3508, opened 2022-11-27
fix(core): detect changes of observable expressions

### LINKED ISSUE #3444: Change detection on components won't work for not default initialized props
**Description**

Creating a bug ticket based on my https://github.com/ngx-formly/ngx-formly/issues/3331#issuecomment-1256092847

Having a component where properties set via observable, and if no default value for that property, it won't fire change detection.

**Minimal Reproduction**

https://stackblitz.com/edit/ngx-formly-ui-bootstrap-pn3yfg

**Your Environment**

- Angular version: 14
- Formly version: 6

**Additional Info**

The observeDeep just hooks on the existing property and that's why it won't pick up any new properties, as the props won't be changes just altered when setting via expressProperties.

aitboudad:
fix #3444

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -121,13 +121,13 @@
                   "with": "integration/ssr/src/environments/environment.prod.ts"
                 }
               ],
-              "optimization": true,
+              "optimization": false,
               "outputHashing": "all",
               "sourceMap": false,
               "namedChunks": false,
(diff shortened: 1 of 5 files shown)


Greegko:
@aitboudad The solution was quite easy it seems. Thanks for the improvement!

PR #3589, opened 2023-02-12
fix(demo): unsubscribe subscriptions to avoid memory leaks

Arooba-git:

Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in SPA applications and their libraries. While running the tool and analyzing the code of ngx-formly, we saw that your project does a very good job of ensuring that all async operations are cancelled when components destroy. However, as per Memlab execution results, we found some unsubscribed subscriptions, causing the memory to leak (screenshots below).

[before]
ngx-formly-before Screen Shot 2023-02-13 at 5 46 05 AM

Hence we added the fix by unsubscribing the subscriptions on component destruction, and you can see the heap size and count of leaks reducing noticeably:
 <br />

after sub Screen Shot 2023-02-13 at 6 14 54 AM

Note. Host listener is used to ensure that ngOnDestroy is called not only when the class destroy, but also when the page refreshes, tab or browser closes or the user navigates away from the page [1]

We executed the test suite after the fixes and it successfully passed 56 of 56 total (in watch + non-watch mode)

You can analyze this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering maximum count of use cases.

Following is a sample of the scenario file we used (it needs to be a .js file but attaching here in .txt form):
[ngx-formly-scenario.txt](https://github.com/ngx-formly/ngx-formly/files/10717563/ngx-formly-scenario.txt)

Note that some other reported leaks (in Memlab) originated from Angular's internal objects, and hence were ignored.

References:
[1] https://wesleygrimes.com/angular/2019/03/29/making-upgrades-to-angular-ngondestroy.html

### DIFF
--- a/demo/src/app/examples/advanced/json-schema/app.component.ts
+++ b/demo/src/app/examples/advanced/json-schema/app.component.ts
@@ -1,15 +1,18 @@
-import { Component } from '@angular/core';
+import { Component, OnDestroy } from '@angular/core';
 import { FormGroup } from '@angular/forms';
 import { FormlyFormOptions, FormlyFieldConfig } from '@ngx-formly/core';
 import { FormlyJsonschema } from '@ngx-formly/core/json-schema';
 import { HttpClient } from '@angular/common/http';
-import { tap } from 'rxjs/operators';
+import { tap, takeUntil } from 'rxjs/operators';
+
+import { Subject } from 'rxjs';
 
 @Component({
   selector: 'formly-app-example',
   templateUrl: './app.component.html',
 })
-export class AppComponent {
+export class AppComponent implements OnDestroy {
+  private destroy$: Subject<any> = new Subject<any>();
   form: FormGroup;
   model: any;
   options: FormlyFormOptions;
@@ -46,11 +49,17 @@ export class AppComponent {
           this.fields = [this.formlyJsonschema.toFieldConfig(schema)];
           this.model = model;
         }),
+        takeUntil(this.destroy$),
       )
       .subscribe();
   }
 
   submit() {
     alert(JSON.stringify(this.model));
+  }
+
+  public ngOnDestroy(): void {
+    this.destroy$.next(true);
+    this.destroy$.complete();
   }
 }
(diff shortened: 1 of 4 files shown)


Arooba-git:
bump @aitboudad

demo/src/app/examples/advanced/json-schema/app.component.ts
```diff
@@ -46,11 +49,18 @@ export class AppComponent {
...
+        takeUntil(this.destroy$),
       )
       .subscribe();
   }
 
   submit() {
     alert(JSON.stringify(this.model));
   }
+
+  @HostListener('unloaded')
```
aitboudad, line 61: Not needed I think

demo/src/app/examples/other/json-powered/app.component.ts
```diff
@@ -40,4 +47,10 @@ export class AppComponent {
       return f;
     });
   }
+
+  @HostListener('unloaded')
```
aitboudad, line 51: same here

demo/src/app/guides/guides.component.ts
```diff
@@ -25,8 +29,14 @@ export class GuidesComponent implements OnInit {
   constructor(private renderer: Renderer2, private route: ActivatedRoute, private elementRef: ElementRef) {}
 
   ngOnInit() {
-    this.route.params.subscribe(({ id }) => {
+    this.route.params.pipe(takeUntil(this.destroy$)).subscribe(({ id }) => {
       this.renderer.setProperty(this.elementRef.nativeElement, 'innerHTML', this.contents[id].default);
     });
   }
+
+  @HostListener('unloaded')
```
aitboudad, line 37: here as well

aitboudad (changes requested):

demo/src/app/examples/advanced/json-schema/app.component.ts
Arooba-git: @aitboudad  updated :)

aitboudad (approved):

aitboudad:
Thank you @Arooba-git!

Issue #3456, opened 2022-10-13
Formly 6 Preset with multiple fields


cagi23:
Hi, 
is possible to have other example of Formly 6 Presets with multiple fields?
The example from the guide explain to configure only one select like this:

```typescript
getConfiguration(): FormlyFieldConfig {
        return {
            key: 'salutation',
            type: 'select',
            props: {
                label: 'Salutation',
                placeholder: 'Please Select',
                options: this.salutationOptions.map((salutation) => ({ label: salutation, value: salutation })),
            },
        };
    }
```

I would configure an array of FormlyFieldConfig and not in the AppModule. 
Is this possible now ?
Thank you so much for the incredible framework!

MaxKless:
Hi @cagi23, 
If I understand you correctly, you want to define a preset with multiple fields? We thought about it but decided against it in #3208.
Let me know if that answers your question :)

cagi23:
Hi @MaxKless,
I'm using https://formly.dev/docs/guide/formly-field-presets/ but I'm not able to configure an array of fields only one single field because the getConfiguration() function return "FormlyFieldConfig" instead I need FormlyFieldConfig[].
Also I need more example of Formly 6's presets in order to better understand the feature.
Thank you!

MaxKless:
Yes that's not supported at the moment because it would add another layer of complexity to the API that we wanted to avoid.
You can define a single field with a field group property to group fields if you'd like!

aitboudad:
If I'm not wrong I think its possible, we decide to let users do the merge by themselves for a such case. So in order to add render multi-fields just use `fieldGroup`.

@cagi23 did you check our example https://formly.dev/docs/examples/other/presets?

cagi23:
Sure, I check your example but I'm a little confused for how implement my reusable set of components.
Specifically, I want to implement something like this :

```typescript
export class AdressPresetProvider {
    constructor(@Inject(ADDRESS_OPTIONS) private adressOptions: string[]) {}
    getConfiguration(): FieldGroup {
        return [ 
          {
            key: 'province',
            type: 'select',
            props: {
                label: 'Province',
                placeholder: 'Please Select',
                options: this.adressOptions.map((adress) => ({ label: address, value: address})),
            },
        },
		{
            key: 'city',
            type: 'select',
            props: {
                label: 'City',
                placeholder: 'Please Select',
                options: this.adressOptions.map((city) => ({ label: city, value: city})),
            },
        }
    }
}
}
[...]

PR #3594, opened 2023-02-14
feat: pass field to FormlyFieldConfigPresetProvider.getConfiguration

MaxKless:
refer to https://github.com/ngx-formly/ngx-formly/issues/3456 for details

@aitboudad the problem is that `reverseDeepMerge` completely ignores arrays, which sucks for the `fieldGroup` merging use case. So I copied and modified it to also handle arrays.

### DIFF
--- a/demo/src/app/app.menu.json
+++ b/demo/src/app/app.menu.json
@@ -114,7 +114,8 @@
           { "path": "/examples/other/button", "label": "Button Type" },
           { "path": "/examples/other/json-powered", "label": "JSON powered" },
           { "path": "/examples/other/input-file", "label": "File input" },
-          { "path": "/examples/other/presets", "label": "Presets" }
+          { "path": "/examples/other/presets", "label": "Presets" },
+          { "path": "/examples/other/presets-fieldgroup", "label": "Fieldgroup Presets" }
         ]
       }
     ]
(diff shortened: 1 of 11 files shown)


aitboudad (changes requested):
I think its better to keep the `reverseDeepMerge`, for two main reasons:

1. It may introduce a breaking change
2. The order of merging an array is relevant for some use-case and that's why I do prefer to let user decide how to merge their data.

MaxKless:
If I keep `reverseDeepMerge` as it is, this example you provided won't work anymore: 
```
export class AdressPresetProvider {
    constructor(@Inject(ADDRESS_OPTIONS) private adressOptions: string[]) {}
    getConfiguration(f: FormlyFieldConfig): FormlyFieldConfig {
        return {
          fieldGroup: [ 
            { key: 'province', ... },
            { key: 'city', ... },
            ...f.fieldGroup,
          ]
        }
    }
}
```

I put some logs into `preset-subsitution.extension.ts` to show what's happening: 
image

you can see that the preset merges the `fieldGroup` correctly but it's ignored by `reverseDeepMerge`. 

Do you have a proposal for how to solve this?

aitboudad:
I see, so it seems a blocking part for now anyway I'll try to further check later and let you know if I find something!

PR #3604, opened 2023-02-25
fix: merge schema properties with definitions $ref

ThibaudAV:

- Remove some duplicated code
- Corrects properties with the use of a definition

to have the same behavior as here  https://rjsf-team.github.io/react-jsonschema-form/

Try with :
```
{
          definitions: {
            address: {
              properties: {
                address1: { type: 'string', title: 'Address 1' },
              },
            },
          },
          type: 'object',
          properties: {
            billing_address: {
              properties: {
                customAddress: { type: 'string', title: 'Custom address' },
              },
              $ref: '#/definitions/address',
            },
          },
        }
```

Only add definitions properties

**What is the new behaviour  (if this is a feature change)?**

Merge add definitions properties and parent one

**Other information**:

### DIFF
--- a/src/core/json-schema/src/formly-json-schema.service.spec.ts
+++ b/src/core/json-schema/src/formly-json-schema.service.spec.ts
@@ -853,6 +853,40 @@ describe('Service: FormlyJsonschema', () => {
         });
       });
 
+      it('should keep schema properties with definition', () => {
+        const schema: JSONSchema7 = {
+          definitions: {
+            address: {
+              properties: {
+                address1: { type: 'string', title: 'Address 1' },
+              },
+            },
+          },
+          type: 'object',
+          properties: {
+            billing_address: {
+              properties: {
+                customAddress: { type: 'string', title: 'Custom address' },
+              },
+              $ref: '#/definitions/address',
+            },
+          },
+        };
+
+        const config = formlyJsonschema.toFieldConfig(schema);
+
+        expect(config.fieldGroup[0]).toEqual({
+          key: 'billing_address',
+          type: 'object',
+          defaultValue: undefined,
+          fieldGroup: expect.any(Array),
+          props: emmptyFieldProps,
+          templateOptions: emmptyFieldProps,
+          validators: expectTypeValidator(['object']),
+        });
+        expect(config.fieldGroup[0].fieldGroup.map((f) => f.key)).toEqual(['customAddress', 'address1']);
+      });
+
       it('should use the locally defined annotations', () => {
         const schema: JSONSchema7 = {
           definitions: {
--- a/src/core/json-schema/src/formly-json-schema.service.ts
+++ b/src/core/json-schema/src/formly-json-schema.service.ts
@@ -423,13 +423,6 @@ export class FormlyJsonschema {
       ];
     }
 
-    if (schema.oneOf && !field.type) {
-      delete field.key;
-      field.fieldGroup = [
-        this.resolveMultiSchema('oneOf', <JSONSchema7[]>schema.oneOf, { ...options, key, shareFormControl: false }),
-      ];
-    }
-
     // map in possible formlyConfig options from the widget property
     if (schema.widget?.formlyConfig) {
       field = this.mergeFields(field, schema.widget.formlyConfig);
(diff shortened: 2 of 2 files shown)


kenisteward:
Based on the Jason schema spec the current implementation is correct.

Definitions should only be mered at the annotation level not at the
validation level.

In order to get what you'd like here you should be using all of instead of
properties definition + ref

kenisteward:
For reference
https://json-schema.org/understanding-json-schema/structuring.html#ref

kenisteward:
It's important for us not to introduce this especially for teams that share
schemas with validators on the backend as those validators would not be
interested this way.

ThibaudAV:
I wanted to test with an allOf but it gives me a result that is not what I expected 😁 

<details>
  <summary>Json Schem</summary>
  
  ```
{
  "type": "object",
  "definitions": {
    "userType": {
      "oneOf": [
        {
          "title": "Human",
          "properties": {
            "pseudo": {
              "type": "string",
              "title": "Pseudo"
            }
          },
          "required": [
            "pseudo"
          ]
        },
        {
          "title": "Robot",
          "properties": {
            "serialNumber": {
              "type": "string",
              "title": "Serial number"
            }
          },
          "required": [
            "serialNumber"
          ]
        }
      ]
    },
    "authMethod": {
      "oneOf": [
        {
          "title": "With token",
          "properties": {
            "token": {
              "type": "string",
              "title": "Token"
            }
          },
          "required": [
            "token"
          ]
        },
        {
          "title": "With username and password",
          "properties": {
            "username": {
              "type": "string",
              "title": "Username"
            },
            "password": {
              "type": "string",
              "title": "Password"
            }
          },
          "required": [
            "username"
          ]
        }
      ]
    }
  },
  "properties": {
    "allOfWithProperties": {
      "title": "AllOf with properties",
      "type": "object",
      "allOf": [
        {
          "properties": {
            "first": {
              "title": "First",
              "type": "string"
            }
          }
        },
        {
          "$ref": "#/definitions/authMethod"
        },
        {
          "properties": {
            "second": {
              "title": "Second",
              "type": "string"
            }
          }
        },
        {
          "$ref": "#/definitions/userType"
        }
      ]
    }
  }
}
```
</details>

I would have expected to have all fiels and multishema in the allOf order. But apparently the AllOf with OneOf not worked as i expect. Only one "OneOf" are displayed and not in the right order.

ThibaudAV:
You can close it ;) 
I hope to find other solutions for my use case.  If you have suggestions, do not hesitate 🙂

kenisteward:
Sorry just saw the updates to this. Could you setup a stackblitz and start
an issue so I can see what output you want vs what output yoy are getting ?

ThibaudAV:
Or course : https://stackblitz.com/edit/angular-jumm2x?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html 

I would expect to have the output fields in the order :
- First field
- authMethod ref fields
- Second field
- userType ref fields

Thank you for your help 🙏

kenisteward:
Ahh yes!   It looks like we might currently have a limitation that we can't
merge disparate oneOf's that are across the same object.

If they are on disparate keys it looks like it works great
https://stackblitz.com/edit/angular-jumm2x-jevk23?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html

If you don't mind, could you make an issue about this that we can track?
It's very possible that we may need to spend some investigation time on the
PR you made here to do the merging this way to alleviate this problem but
there may be some other solution as well.  We also could add an option to
the formly field generate that is something like "mergeReferences" that
would allow this to work for the normal use case

what do you think? @aitboudad?

aitboudad:
I'll revised this once time allows :pray:, have you tried adjusting the order through the map option https://github.com/ngx-formly/ngx-formly/issues/1982#issuecomment-566048527

kenisteward:
@ThibaudAV what @aitboudad does fix the ordering issue!  I forgot this was added in order to define what order the fields which fixed that.

The other issue of logic for merging oneOf's seems like it's a bigger discussion ticket though.

In the context of validations it makes sense because jsonschema is just a validation language that parses them disparately as separate rules. When it comes to generating fields though we have to decide if we merge them what exactly happens.  Right now it looks like first one wins ( or last one I can't remember haha )

ThibaudAV:
@kenisteward 
> If they are on disparate keys it looks like it works great
https://stackblitz.com/edit/angular-jumm2x-jevk23?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html

This changes the structure of the object and I can't change that 

No the order does not fix my problem :/ 

And yes for OneOf merger. Maybe there is something to change on this side. maybe add them in a respective sub form. or something like that

Issue #3616, opened 2023-03-07
Classname has no effect on FormlyFieldType using formly 6 and material 15


christophglass:
**Description**
Classname has no effect on FormlyFieldType using Angular Material 15

**Expected behaviour**
`<mat-slide-toggle class="mat-mdc-slide-toggle mat-accent ng-pristine ng-valid ng-touched slider-full-width-before"...></mat-slide-toggle>`

**Actual behaviour**
`<mat-slide-toggle class="mat-mdc-slide-toggle mat-accent ng-pristine ng-valid ng-touched "...></mat-slide-toggle>`

**Minimal Reproduction**
```
 {
    key: 'decision',
    type: 'toggle',            
    className: 'slider-full-width-before',
    wrappers: ['form-field', 'togglewrapper'],
    props: {
      labelPosition: 'before',
      label: 'yes or no',
      description: 'make a decision please',
    },
  }
```

Classname doesnt show up in DOM. If I add class in Chrome devtools it works.

**Your Environment**

- Angular version: 15.2.0
- Formly version: 6.1.1
- Material: 15.2.0

**Additional context**
Class 'slider-full-width-before' comes from custom material theme.
Yes I even put a custom wrapper around it hoping I could add the missing class somehow after view init.

christophglass:
**Workaround using wrapper**

```
import { AfterViewInit, Component, ViewChild, ViewContainerRef } from '@angular/core';
import { FieldWrapper } from '@ngx-formly/core';

@Component({
  selector: 'ng-online-calculator-ui-toggle-wrapper',
  templateUrl: './toggle-wrapper.component.html',
  styleUrls: ['./toggle-wrapper.component.css']
})
export class ToggleWrapperComponent extends FieldWrapper implements AfterViewInit {  
  @ViewChild('fieldComponent', { read: ViewContainerRef }) declare fieldComponent: ViewContainerRef;

  ngAfterViewInit(): void {    
    document.getElementById(this.fieldComponent.element.nativeElement.previousElementSibling.firstChild.id)?.classList.add('slider-full-width-before');        
  }
}

```

Dont know if this is a good "solution" but it works at least...

aitboudad:
The `className` is added to `formly-fieled` element which is located at the root of `toggle` field type, to update the style of `mat-slide-toggle` element just use the following selector:

```css
.slider-full-width-before .mat-mdc-slide-toggle {
  ...
}
```

PR #3634, opened 2023-03-19
feat(material): add panelClass to mat-select props

jazpearson:
Feature

Cannot style the mat-options when using a formly select. Is there another way to do this with the current behaviour? 

Allows you to specify a class name to the mat-options. This enables the user to help theme the mat-select control. 

**Other information**:

### DIFF
--- a/src/ui/material/select/src/select.type.spec.ts
+++ b/src/ui/material/select/src/select.type.spec.ts
@@ -124,7 +124,7 @@ describe('ui-material: Formly Field Select Component', () => {
       expect(queryAll('mat-option')).toHaveLength(3);
     });
 
-    it('mat-options should have specified custom class', () => {
+    it('panelClass should be used to add class to select overlay div', () => {
       const panelClass = 'my-custom-panel-class';
 
       const { query, queryAll, detectChanges } = renderComponent({
--- a/src/ui/material/select/src/select.type.spec.ts
+++ b/src/ui/material/select/src/select.type.spec.ts
@@ -123,6 +123,35 @@ describe('ui-material: Formly Field Select Component', () => {
 
       expect(queryAll('mat-option')).toHaveLength(3);
     });
+
+    it('mat-options should have specified custom class', () => {
+      const optionsClass = 'my-custom-class';
+
+      const { query, queryAll, detectChanges } = renderComponent({
+        key: 'sportId',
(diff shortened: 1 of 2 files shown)


aitboudad:
Hi, thanks for the PR!
I think you can style `options` through `className` property, you may check my comment hope it may help https://github.com/ngx-formly/ngx-formly/issues/3616#issuecomment-1458685515

jazpearson:
As far as i can tell, the `className` gets added to the parent `mat-select`. But in the resulting overlay, nothing seems to be passed down to the `mat-option`s.  

But i'm new to this code-base so will take a look. Thanks for the very fast response!

jazpearson:
And i want to be able to dynmically style the mat-options so i can switch between multiple themes.

jazpearson:
@aitboudad I've tried adding the `className` but it doesn't get passed down to the mat-options. The mat-options get generated inside a cdk overlay at run-time.... so it looks like you have to style the `mat-option` itself. 

Any suggestions?

https://stackblitz.com/edit/ngx-formly-select-value-text-ntapbk?file=src/styles.scss

aitboudad:
I see, so instead I suggest introducing a new property: `panelClass` to `mat-select` (https://material.angular.io/components/select/api#MatSelect)

jazpearson:
That's one of the first things i tried. I can get things working with `panelClass` if i hand-write the HTML, but if i use formly, `panelClass` doesn't seem to work. 

Using formly - and if you check the generated HTML, you can see the `panelclass` attribute:
https://stackblitz.com/edit/ngx-formly-select-value-text-subh7a?file=src/app/app.component.ts

Hand-written HTML example:
https://stackblitz.com/edit/angular-dua1of?file=src/styles.scss

I also found the same question being asked on stack overflow, but unanswered: https://stackoverflow.com/questions/75554182/angular-material-15-how-can-i-apply-the-panelclass-property-to-mat-select-u

aitboudad:
I understand, please update this PR to add `panelClass` input to `mat-select` component.

jazpearson:
@aitboudad updated. and seems to work :)

aitboudad:
This PR has been released as part of [v6.1.4](https://github.com/ngx-formly/ngx-formly/releases/tag/v6.1.4) release.

Thank you!

PR #3671, opened 2023-05-04
fix(core): handle undefined source observer

gkamperis:
Bug fix

After upgrading to v6 there were certain scenarios (mainly in tests of our system) where reassigning the full config many times in succession would result in the `source` of the utility function `observeDeep` to be called with `undefined`
This would result in an invalid unsubscribe function to be returned.

`source` is defined as `any` but the current code does not cater for handling the "falsy" scenario

When `source` is `undefined` or `null` an empty `unsubscribe` function is returned allowing callers to not crash.

### DIFF
--- a/src/core/src/lib/utils.ts
+++ b/src/core/src/lib/utils.ts
@@ -225,7 +225,7 @@ interface IObserveTarget<T> {
   };
 }
 
-export function observeDeep(source: any, paths: string[], setFn: () => void): () => void {
+export function observeDeep<T = any>(source: IObserveTarget<T>, paths: string[], setFn: () => void): () => void {
   let observers: Function[] = [];
 
   const unsubscribe = () => {
@@ -250,6 +250,10 @@ export function observeDeep(source: any, paths: string[], setFn: () => void): ()
 }
 
 export function observe<T = any>(o: IObserveTarget<T>, paths: string[], setFn: IObserveFn<T>): IObserver<T> {
+  if (!isObject(o)) {
+    o = {};
+  }
+
   if (!o._observers) {
     defineHiddenProp(o, '_observers', {});
   }
--- a/src/core/src/lib/components/formly.field.ts
+++ b/src/core/src/lib/components/formly.field.ts
@@ -268,7 +268,7 @@ export class FormlyField implements DoCheck, OnInit, OnChanges, AfterContentInit
 
     const subscribes = [
       observeDeep(field, ['props'], () => field.options.detectChanges(field)),
-      observeDeep(field.options, ['formState'], () => field.options.detectChanges(field)),
+      observeDeep(field, ['options', 'formState'], () => field.options.detectChanges(field)),
     ];
 
     for (const key of Object.keys(field._expressions || {})) {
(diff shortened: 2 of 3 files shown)


gkamperis:
@aitboudad can you pls check this? fix is required for v6 upgrade.

src/core/src/lib/utils.ts
```diff
@@ -226,6 +226,10 @@ interface IObserveTarget<T> {
 }
 
 export function observeDeep(source: any, paths: string[], setFn: () => void): () => void {
+  if (source == null) {
```
aitboudad, line 229: There is no issue in `observeDeep`, I think that check should be handled in `observe` function instead

gkamperis: the issue is in `observeDeep` for sure.  
it is passed an undefined `source` which is passed to the `observe` and then when `unsubscribe` is called it NPEs on the if() check of `observe()`.

`observeDeep` takes `any` so code has to check for falsy values.
The fix results in a no-op unsubscribe and fixes the issue.

aitboudad:
Thank you @gkamperis!

src/core/src/lib/utils.ts
gkamperis: @aitboudad I see you updated the PR and in essence completely removed my fix.

Although you are now handling `undefined` by creating an object the signature is still incorrect and misleading.

either `observe()` is always passed in an object or it should become `source: IObserveTarget<T> | undefined`

if not it is misleading and then this does not make sense  `if (!isObject(o)) {` as it should never happen according to the signature and TS.

Also with your fix you are still creating an empty observer which will no-op if source is indeed `undefined`. 
Why do that?

aitboudad: @gkamperis `observeDeep` rely on `observe` so I think the fix should be handled inside `observe`. The solution is either to raise an error in case of passing an empty object or handle it somehow.

Could you please share the original error that produce passing an empty value to `observeDeep` so I can see if the current fix can be handled by other ways?

aitboudad: FYI I've reverted this PR and instead added a check to ensure that passed value to `observeDeep` is not empty.

gkamperis: @aitboudad the NPE was coming from `observeDeep(field.options, ` 
so `field.options` was `undefined`

I see you changed it so `options` is now a `path`. 
not sure how the other changes in your latest PR are related and if they will stop the original issue.

tbh idk enough about the usage of the function across the project.

since `observeDeep` is a util function for me it would be best to handle the issue within it.
you now have to make sure all callers pass a valid object.

but still `private fieldChanges(field: FormlyFieldConfigCache) {`
is contradicting with `  if (!field)  {;`

it should be `private fieldChanges(field: FormlyFieldConfigCache | undefined) {`

aitboudad: > not sure how the other changes in your latest PR are related and if they will stop the original issue.

other changes not related to this.

> tbh idk enough about the usage of the function across the project.

For now only used inside `FormlyField` component

> it should be private fieldChanges(field: FormlyFieldConfigCache | undefined) {

Right, pushed a fix.

aitboudad: The fix is published in https://github.com/ngx-formly/ngx-formly/releases/tag/v6.1.5

PR #3792, opened 2023-10-05
perf(core): set the asyncValidators when async validators are present

### LINKED ISSUE #3791: Adding item to large array takes very long
**Description**
When the length of array increases, Add operation on array takes quite a long. In my case it is taking 4 sec.

**Minimal Reproduction**
Create an array which item are nested at-least 2 level and with at least 9 to 10 fields. Now start adding items to array, will observe the amount of time it take to perform add operation increases. 

Formly version: 6.1.8

Additional context:

I found that building array fields is taking a lot of time. Reason is for all the array items and subfields **"extension.postPopulate"** is called, which is further calling the **"setValidators"** method of **"field-form.ts"** here the creation of observable for async validations takes a lot of time.

```
c.setAsyncValidators(() => {
            const v = Validators.composeAsync(this.mergeValidators<AsyncValidatorFn>(field, '_asyncValidators'));
            return v ? v(c) : of(null);
    });
``` 

I did the following POC which reduced this time from **4sec to 0.6sec**.

**What I did**
I added an if condition before the "c.setAsyncValidators" method to check if we have async validators.
```
const asyncValidatorsList =
    this.mergeValidators<AsyncValidatorFn>(field, '_asyncValidators');

asyncValidatorsList.length && c.setAsyncValidators(() => {
     const v = Validators.composeAsync(asynValidatorsList);
     return v ? v(c) : of(null);
   });
```

dev-kanishk:
The array add operation was slow because the field build created an observable and set the asyncValidators, even when there were no async validators. This change checks if the async validators are present and then only sets them.

fix #3791

Improves the performance of array add operation, mentioned in the issue #3791

### DIFF
--- a/src/core/src/lib/extensions/field-form/field-form.ts
+++ b/src/core/src/lib/extensions/field-form/field-form.ts
@@ -98,21 +98,24 @@ export class FieldFormExtension implements FormlyExtension {
         }
 
         if (null === c.validator) {
-          c.setValidators(() => {
-            const v = Validators.compose(this.mergeValidators<ValidatorFn>(field, '_validators'));
+          const validatorsList = this.mergeValidators<ValidatorFn>(field, '_validators', true);
 
-            return v ? v(c) : null;
-          });
+          if (validatorsList.length) {
+            c.setValidators(() => {
+              const v = Validators.compose(this.mergeValidators<ValidatorFn>(field, '_validators'));
 
-          markForCheck = true;
+              return v ? v(c) : null;
+            });
+            markForCheck = true;
+          }
         }
 
         if (null === c.asyncValidator) {
-          const asyncValidatorsList = this.mergeValidators<AsyncValidatorFn>(field, '_asyncValidators');
+          const asyncValidatorsList = this.mergeValidators<AsyncValidatorFn>(field, '_asyncValidators', true);
 
           if (asyncValidatorsList.length) {
             c.setAsyncValidators(() => {
-              const v = Validators.composeAsync(asyncValidatorsList);
+              const v = Validators.composeAsync(this.mergeValidators<AsyncValidatorFn>(field, '_asyncValidators'));
               return v ? v(c) : of(null);
             });
             markForCheck = true;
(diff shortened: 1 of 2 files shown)


src/core/src/lib/extensions/field-form/field-form.ts
```diff
@@ -97,21 +97,28 @@ export class FieldFormExtension implements FormlyExtension {
...
-            const v = Validators.composeAsync(this.mergeValidators<AsyncValidatorFn>(field, '_asyncValidators'));
-
-            return v ? v(c) : of(null);
-          });
 
           markForCheck = true;
         }
 
+        if (null === c.asyncValidator) {
+          const asyncValidatorsList = this.mergeValidators<AsyncValidatorFn>(field, '_asyncValidators');
```
aitboudad, line 111: This will break some cases if I'm not wrong, the main reason of calling `mergeValidators` inside `setAsyncValidators` is to keep validation in sync with field changes.

dev-kanishk: If I am understanding correct is that you are saying the the list of asyncValidators for a field can change without the need of rebuilding the field.

aitboudad: At runtime I think, when the field is hidden for example the validation should be updated. Try to apply the same changes to `setValidators`, and you'll notice some test failures.

dev-kanishk: I have taken the account of hidden case. And all the test cases are passing.

aitboudad (approved):

aitboudad:
Thank you @dev-kanishk!

dev-kanishk:
Your welcome @aitboudad, Thank you for accepting the change. It improves my project array add operation by 4X.
8dUCSi7bFAfuCgA

aitboudad:
@dev-kanishk The perf gain is significant, many thanks!
Could you please check the perf gain of another improvement: https://github.com/ngx-formly/ngx-formly/commit/3f65fb482e0d9eb89d573e2d54cc650b1fd9eddf ?

dev-kanishk:
Yeah, here is the perf gain, its of around 1.08
BnjsXHbvQ6H5dEJ
[Formly performance improvement - Sheet2 (1).pdf](https://github.com/ngx-formly/ngx-formly/files/12839754/Formly.performance.improvement.-.Sheet2.1.pdf)

PR #4021, opened 2024-11-20
fix(testing): mark components as standalone false

### LINKED ISSUE #4020: createFieldComponent broken in Angular 19
**Description**
In Angular 19, [standalone is the new default](https://blog.angular.dev/the-future-is-standalone-475d7edbc706). When using the `createFieldComponent()` function from `@angular/core/testing` in an Angular 19 project, the following error is thrown:

>Error: Unexpected "TestComponent" found in the "declarations" array of the "TestBed.configureTestingModule" call, "TestComponent" is marked as standalone and can't be declared in any NgModule - did you intend to import it instead (by adding it to the "imports" array)?

The reason is that with Angular 19 this component is now standalone. However the tooling expects a non-standalone component, so we need to explicitly mark this component as `standalone: false`.

**Minimal Reproduction**

I can create a minimal reproduction if necessary, but I think the issue is clear.

**Your Environment**

- Angular version: 19.0.0
- Formly version: 6.3.10

**Additional context**
n/a

json-derulo:
> Note: Currently this changes produce build errors, as the workspace is still on Angular 13 which doesn't support the `standalone` component metadata just yet

Bug fix

Closes #4020

The components in the testing entry point are explicitly marked as `standalone: false`, so they work in Angular 19.

**Other information**:

### DIFF
--- a/src/core/testing/src/component-factory.ts
+++ b/src/core/testing/src/component-factory.ts
@@ -112,6 +112,7 @@ export function createFieldComponent(
   selector: 'formly-test-component',
   template: '',
   providers: [FormlyFormBuilder],
+  standalone: false,
 })
 class TestComponent {
   constructor(public builder?: FormlyFormBuilder) {}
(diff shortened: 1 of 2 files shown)


aitboudad:
Thank you @json-derulo!

PR #4093, opened 2025-05-25
refactor(core+material): convert to built-in control flow

atsjo:
Refactor to use built-in control flow for core and material, which was made stable in angular v18

Uses *ngFor/*ngIf style control flow, which will be deprecated in angular v20

No change in behavior, except probably better performance

**Other information**:


kenisteward:
@aitboudad 

Will this just be a merge for v7 only? I think if we release in v6 it will break anyone using lower than 17 for their app right?

atsjo:
It was a PR against next, which depends on ^18 . Build would fail against main, which depends on ^13

kenisteward:
Ahh I missed that because I'm on my phone. That is perfect!

PR #4179, opened 2026-05-04
fix(json-schema): support recursive multischema branches

### LINKED ISSUE #3989: Maximum call stack size exceeded with self referencing JSON Schema models
**Description**

When using JSON Schema that contains recursive schema references it often happens that we get a "Maximum call stack size exceeded" error

**Minimal Reproduction**

Here is a minimal example to reproduce the problem :

https://stackblitz.com/edit/stackblitz-starters-aag9bp?file=tsconfig.json,src%2Fassets%2Fjson-schema%2Frecurse.json

The condition property of the rule object is a `oneOf` that list all the boolean operation possibles. The `And`, `Or` and `Not` operations contain themselves a predicate property which is also a `oneOf` listing all the boolean operations, hence themselves.

When one click on the add button, a  "Maximum call stack size exceeded" error is thrown.

```
.localservice@runtime.e7ca9335.js:26 ERROR RangeError: Maximum call stack size exceeded
    at FormlyJsonschema._toFieldConfig (vendor.js:90637:17)
    at vendor.js:91082:19
    at Array.map (<anonymous>)
    at FormlyJsonschema.resolveMultiSchema (vendor.js:91081:29)
    at FormlyJsonschema._toFieldConfig (vendor.js:91001:32)
    at vendor.js:90816:28
    at Array.forEach (<anonymous>)
    at FormlyJsonschema._toFieldConfig (vendor.js:90814:48)
    at vendor.js:91082:19
    at Array.map (<anonymous>)
```

**Your Environment**

- Angular version: 16 (but also occurs with Angular 18)
- Formly version: 6.3.7

OlimjonovOtabek:
## Summary

- detect recursive JSON Schema `oneOf` / `anyOf` multischema branches
- avoid eagerly expanding recursive branches during initial field generation
- lazily build recursive branches only when they become selected
- add regression coverage for recursive `oneOf` branches

## Root Cause

`resolveMultiSchema()` eagerly mapped every branch with `_toFieldConfig()`. For self-referencing schemas, a branch can resolve back to the same multischema definition, causing `_toFieldConfig()` and `resolveMultiSchema()` to recurse until the stack overflows.

## Behavior

Non-recursive multischema schemas keep the existing eager behavior. Recursive multischema schemas use a lazy branch expansion path and skip nested multischema auto-selection during validation probes, preventing the stack overflow while still allowing selected recursive branches to render.

## Validation

- `npx jest src/core/json-schema/src/formly-json-schema.service.spec.ts --runInBand`
- `npx prettier --check src/core/json-schema/src/formly-json-schema.service.ts src/core/json-schema/src/formly-json-schema.service.spec.ts`
- `npx eslint src/core/json-schema/src/formly-json-schema.service.ts src/core/json-schema/src/formly-json-schema.service.spec.ts` *(passes with existing warnings)*
- `git diff --check`

Fixes #3989

### DIFF
--- a/src/core/json-schema/src/formly-json-schema.service.spec.ts
+++ b/src/core/json-schema/src/formly-json-schema.service.spec.ts
@@ -1581,6 +1581,50 @@ describe('Service: FormlyJsonschema', () => {
           expect(field.model).toEqual({ foo: 2 });
         });
 
+        it('should support recursive oneOf schema branches', () => {
+          const { field, detectChanges } = renderComponent({
+            schema: {
+              definitions: {
+                predicate: {
+                  oneOf: [
+                    {
+                      title: 'Equals',
+                      type: 'object',
+                      properties: {
+                        value: { type: 'string' },
+                      },
+                    },
+                    {
+                      title: 'Not',
+                      type: 'object',
+                      properties: {
+                        predicate: { $ref: '#/definitions/predicate' },
+                      },
+                    },
+                  ],
+                },
+              },
+              type: 'object',
+              properties: {
+                condition: { $ref: '#/definitions/predicate' },
+              },
+            },
+          });
+
+          const [
+            conditionSelector,
+            {
+              fieldGroup: [, notField],
+            },
+          ] = field.fieldGroup[0].fieldGroup[0].fieldGroup;
+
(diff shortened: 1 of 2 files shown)


aitboudad:
Thank you @OlimjonovOtabek!

PR #4184, opened 2026-05-18
fix(core): detect select option observables with isObservable

### LINKED ISSUE #4173: FormlySelectOptionsPipe incorrectly relies on instanceof Observable
## Summary

`FormlySelectOptionsPipe` relies on `instanceof Observable` to detect streams.  
This check fails when the Observable originates from a different RxJS instance.

As a result, valid Observables may be treated as static objects, causing Formly to render incorrect options.

## Description

Currently, `FormlySelectOptionsPipe` uses `instanceof Observable` to determine whether the `options` property is a stream or a static array.

While this works when a single RxJS instance is used, it breaks in several modern scenarios where multiple RxJS instances may exist.

Examples include:
- multiple `rxjs` copies resolved in `node_modules`
- micro-frontends / Module Federation
- multi-version dependency trees during migrations

In these cases the value is still a valid Observable (it has a `.subscribe()` method), but `instanceof Observable` returns `false`.

## Minimal reproduction

StackBlitz reproduction:

https://stackblitz.com/edit/stackblitz-starters-bszajeus?file=src%2Fmain.ts

The example contains two Formly fields:

- **Field A:** options coming from an **Elf store** (`Observable<SelectOption[]>`)
- **Field B:** options fetched **directly via HTTP** (`Observable<SelectOption[]>`)

Both values are valid Observables.

However, Formly correctly processes the HTTP Observable but fails to detect the store Observable.

## Steps to reproduce

1. Open the StackBlitz reproduction  
2. Open the browser **DevTools console**  
3. Observe the console output
The reproduction logs the following comparison:

| source | instanceof Observable | isObservable |
|------|------|------|
| Elf store | false | true |
| Direct HTTP | true | true |

This shows that the store stream is a valid Observable (`isObservable === true`) but fails the `instanceof Observable` check.

## Expected behavior

Formly should recognize the Observable stream, subscribe to it, and render the emitted options.

## Actual behavior

Because the `instanceof Observable` check fails, the pipe treats the Observable instance as a plain object (static data).

Since Formly expects an array of options, it attempts to map this object into `FormlyFieldSelectOption`.

Because the Observable instance does not contain properties like `label` or `value`, the mapping falls back to its default state.

As a result, the pipe returns something like:
```
[{ disabled: false }]
```

In the UI this appears as a **single empty option**, instead of rendering the actual options from the stream.

This results in a **silent failure**, which makes the issue difficult to diagnose.

## Screenshot

In the screenshot below, the field whose options come from the store does not render any options, while the HTTP-based field works correctly.

Screenshot

## Code reference
https://github.com/ngx-formly/ngx-formly/blob/09f891c9b7473cea365b7bcf2d1cad2de1a1846a/src/core/select/src/select-options.pipe.ts#L33-L37

## Suggested fix

Use `isObservable` from `rxjs`, which correctly detects Observables across different RxJS instances.
```ts
import { isObservable } from 'rxjs';

if (!isObservable(options)) {
  options = this.observableOf(options, field);
} else {
  this.dispose();
}
```
Alternatively, a simple duck-typing check would also work:

```ts
function isObservable(obj: any): obj is Observable<any> {
  return !!obj && typeof obj.subscribe === 'function';
}
```

## Environment

The issue occurs in a real project with the following environment:
```
Angular: 19.2.x
@ngx-formly/core: 6.3.12
@ngneat/elf: 2.5.x
rxjs: 7.8.x
```

## Reproduction environment

The minimal reproduction uses the latest versions available in StackBlitz and the issue is still reproducible:
```
Angular: 21.1.x
@ngx-formly/core: 7.1.x
@ngneat/elf: 2.5.x
rxjs: 7.8.x
```

This suggests the issue is not tied to a specific Angular or Formly version and is likely related specifically to the use of `instanceof Observable`.

## Additional context

The issue was discovered while integrating Formly with **@ngneat/elf** state management.

The store stream behaves correctly with RxJS operators and `isObservable`, but fails the `instanceof Observable` check used internally by Formly.

sjh9714:
## Summary

Fixes select option Observable detection across different RxJS instances.

## What changed

- Added regression coverage for an Observable-like value that is not an instance of the local RxJS `Observable` class.
- Updated `FormlySelectOptionsPipe` to use RxJS `isObservable` instead of `instanceof Observable`.

## Why

Issue #4173 reports that valid Observables from another RxJS instance can fail the `instanceof Observable` check and get treated as static options.

## Testing

- `npx jest src/core/select/src/select-options.pipe.spec.ts --runInBand`
- `npm run lint -- --quiet`
- `npm test -- --runInBand --silent`
- `git diff --check HEAD~1..HEAD`

Fixes #4173

### DIFF
--- a/src/core/select/src/select-options.pipe.spec.ts
+++ b/src/core/select/src/select-options.pipe.spec.ts
@@ -1,5 +1,5 @@
 import { FormlySelectOptionsPipe } from './select-options.pipe';
-import { of as observableOf, Subject } from 'rxjs';
+import { firstValueFrom, isObservable, Observable, of as observableOf, Subject } from 'rxjs';
 import { FormlyFieldConfig } from '@ngx-formly/core';
 import { FormlyValueChangeEvent } from 'src/core/src/lib/models';
 
@@ -26,6 +26,22 @@ describe('Pipe: FormlySelectOptionsPipe', () => {
     pipe.transform(observableOf([{ label: '1', value: '1' }])).subscribe((options) => {
       expect(options).toEqual([{ label: '1', value: '1', disabled: false }]);
     });
+  });
+
+  it('passing options as an observable that is not an Observable instance', async () => {
+    const source = observableOf([{ label: '1', value: '1' }]);
+    const crossInstanceObservable = {
+      lift: source.lift.bind(source),
+      pipe: source.pipe.bind(source),
+      subscribe: source.subscribe.bind(source),
+    };
+
+    expect(crossInstanceObservable instanceof Observable).toBe(false);
+    expect(isObservable(crossInstanceObservable)).toBe(true);
+
+    await expect(firstValueFrom(pipe.transform(crossInstanceObservable))).resolves.toEqual([
+      { label: '1', value: '1', disabled: false },
+    ]);
   });
 
   it('should detect field expressions changes', () => {
--- a/src/core/select/src/select-options.pipe.ts
+++ b/src/core/select/src/select-options.pipe.ts
@@ -1,5 +1,5 @@
 import { OnDestroy, Pipe, PipeTransform } from '@angular/core';
-import { BehaviorSubject, Observable, Subscription } from 'rxjs';
+import { BehaviorSubject, isObservable, Observable, Subscription } from 'rxjs';
 import { filter, map, tap } from 'rxjs/operators';
 import { FormlyFieldConfig, FormlyFieldProps } from '@ngx-formly/core';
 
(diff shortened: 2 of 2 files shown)


dmytro-develop:
Thanks for picking this up.

I noticed there is a similar `instanceof Observable` check in `field-expression-legacy/field-expression.ts`:

https://github.com/ngx-formly/ngx-formly/blob/09f891c9b7473cea365b7bcf2d1cad2de1a1846a/src/core/src/lib/extensions/field-expression-legacy/field-expression.ts#L46

Since this seems to have the same cross-RxJS-instance issue, would it make sense to replace it with `isObservable(expr)` in this PR as well?

`isObservable` is already imported in that file and used later in `parseExpressions`, so the actual change should be very small. I can also open a separate issue/PR for this if you prefer to keep this PR scoped only to `FormlySelectOptionsPipe`.

sjh9714:
Thanks for pointing that out. I agree it looks like the same pattern.

I kept this PR focused on the select options path from #4173, but I’m happy to update this PR if maintainers prefer covering the legacy field-expression path here as well. Otherwise, a small follow-up PR would probably keep the scope clearer.

aitboudad (approved):

aitboudad:
Thank you @sjh9714!

